PR #5, opened 2017-11-30
Make TLE schema and Survey metadata part of the extension files

whiskeyjay:
Background: the extension used read from a redirection file stored in one of our Azure Blob containers which points to the actual location of the TLE file, for versioning purpose and made it possible to update the TLE file without having to republish the extension.

This change makes the TLE/Survey metadata files part of the extension files. It means we will have to publish a new version of the extension each time the TLE gets updated. This shouldn't be a problem given how VS Code updates its extensions.

image


src/AzureRMAssets.ts
```diff
@@ -73,83 +73,41 @@ export class AzureRMAssets {
      */
     public static getFunctionMetadataUri(): Promise<string> {
         if (AzureRMAssets._functionMetadataUri === undefined) {
-            AzureRMAssets._functionMetadataUri = AzureRMAssets.getAssetUri("ExpressionMetadata.json");
+            AzureRMAssets._functionMetadataUri = AzureRMAssets.getLocalAssetUri("ExpressionMetadata.json"); //AzureRMAssets.getAssetUri("ExpressionMetadata.json");
```
EricJizbaMSFT: Can we get rid of the commented code?

src/AzureRMAssets.ts
```diff
@@ -73,83 +73,41 @@ export class AzureRMAssets {
...
-    private static getAssetUri(assetFileName: string): Promise<string> {
-        return AzureRMAssets.getCurrentVersionRedirectUri()
-            .then((versionRedirectUri: string) => {
-                return versionRedirectUri + (versionRedirectUri.endsWith("/") ? "" : "/") + assetFileName;
-            });
+    private static getLocalAssetUri(assetFileName: string): Promise<string> {
+        return Promise.resolve<string>(path.join(__filename, "..", "..", "..", "assets", assetFileName));
+    }
+
+    private static readFile(path: string): Promise<string> {
```
EricJizbaMSFT: Not necessary for this PR, but I'd take a look at fs-extra. It's got promisified versions of all the fs functions (plus some other useful stuff)

PR #21, opened 2018-02-01
Fix #20

### LINKED ISSUE #20: Cannot read properties of null
The TreeView throws an exception "Cannot read properties of null" when the JSON document is empty.

whiskeyjay:
Fix #20 
Sorry about the messy diff, VS Code formatted the code.

The only real change is at new line 43.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
                 {
                     "id": "json-outline",
                     "name": "JSON Outline",
-                    "when": "showArmJsonView"
+                    "when": "resourceLangId == json"
                 }
             ]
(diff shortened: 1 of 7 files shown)


marcvaneijk:
np for the diff. Thanks for your work on reviewing the PR. 

The refresh method is triggered when the provider is initialized or a new .json file is opened with the provider already initialized, in both cases clearing the tree (with the parseTree method). So isn't the tree empty at all times when no element is passed to the getChildren method? (as the element is only passed when expanding a root element in the tree).

PR #22, opened 2018-02-01
Make JSON outline show only for ARM templates

### LINKED ISSUE #17: JSON outline is shown for any JSON files
The current design of the JSON outline TreeView is specific for ARM templates, for example, the icons. The TreeView should only appear for ARM templates.

### LINKED ISSUE #18: JSON outline not shown for unsaved new ARM template
1. In VS Code, create a new document.
2. Change the document type to JSON from the lower right corner.
3. Type or paste in some valid JSON content.

Expected:
The JSON outline tree is shown.

Actual:
The JSON outline tree is not shown.

whiskeyjay:
Fix #17 - Now the extension detects for ARM template schema in a JSON, only when an ARM template is detected, the TreeView will be visible.

Fix #18 - Now the TreeView will be visible for newly created but not yet saved to disk ARM templates.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
                 {
                     "id": "json-outline",
                     "name": "JSON Outline",
-                    "when": "resourceLangId == json"
+                    "when": "showArmJsonView"
                 }
             ]
         }
--- a/src/Treeview.ts
+++ b/src/Treeview.ts
@@ -1,7 +1,8 @@
 import * as Json from "./JSON";
 import * as vscode from "vscode";
 import { TreeItem } from "vscode";
-
+import { ContextTagKeys } from "applicationinsights/out/Declarations/Contracts";
+import * as Utilities from "./Utilities";
 
 export class JsonOutlineProvider implements vscode.TreeDataProvider<string> {
     private tree;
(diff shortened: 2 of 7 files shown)


src/Treeview.ts
```diff
@@ -297,6 +292,40 @@ export class JsonOutlineProvider implements vscode.TreeDataProvider<string> {
...
+    private isArmTemplate(document?: vscode.TextDocument): boolean {
+        return !!document && document.languageId.toLowerCase() === 'json' && Utilities.isValidSchemaUri(this.getSchemaUri());
+    }
+
+    private setTreeViewContext(visible: boolean) {
+        vscode.commands.executeCommand('setContext', 'showArmJsonView', visible);
+    }
+
+    private getSchemaUri(): string {
+        if (!!this.tree) {
```
EricJizbaMSFT: Can any of these checks throw an exception? For example, `Json.asObjectValue` or `Json.asStringValue` or `schema.toString()`?

If it's remotely possible, I'd prefer to wrap this in a try/catch and ignore all errors since we perform this check for every single json file (correct?)

src/Treeview.ts
```diff
@@ -297,6 +292,40 @@ export class JsonOutlineProvider implements vscode.TreeDataProvider<string> {
...
+
+        if (showTreeView) {
+            this.refresh();
+        }
+
+        this.setTreeViewContext(showTreeView);
+    }
+
+    private isArmTemplate(document?: vscode.TextDocument): boolean {
+        return !!document && document.languageId.toLowerCase() === 'json' && Utilities.isValidSchemaUri(this.getSchemaUri());
```
EricJizbaMSFT: Does getting the schema uri have any performance impact on large json files? In other words - does it parse the whole file and would the user notice any delay in their non-ARM json files because of the ARM extension?

package.json
```diff
@@ -44,7 +44,7 @@
                 {
                     "id": "json-outline",
                     "name": "JSON Outline",
```
EricJizbaMSFT: nit: Since these only apply to ARM - can ARM be in the title of the explorer? "JSON Outline" seems too generic
screen shot 2018-02-01 at 10 19 16 am

src/Treeview.ts
whiskeyjay: At this point the whole file has already been parsed. Getting the schema doesn't do a re-parse. There are room for perf improvements, but that will require much changes. I'd prefer not to do it at this time.

src/Treeview.ts
whiskeyjay: No, these functions shouldn't throw exceptions. If the values can't be converted, they return null, which code does the check.

PR #40, opened 2018-03-12
Add upload-vsix to travis ci and extension recommendations

EricJizbaMSFT:
While I was at it, I cleaned up some other aspects of the repo. There's definitely more work to be done, but I just took care of some low-hanging fruit:
1. Move changelog to its own file
1. Change README sections to be the same as our other extensions
1. Remove some 'cmd' files that are no longer needed
1. Mark this extension as 'preview'


.travis.yml
```diff
@@ -7,9 +7,10 @@ install:
   - npm install
 
 script:
-  - npm run vscode:prepublish
+  - gulp package
```
StephenWeatherford: Is there any difference between this and running "vsce package" from command line?

EricJizbaMSFT: `vsce package` requires you to have vsce installed globally on your machine. If you we do it through gulp, we can specify 'vsce' as a devDependency in package.json and it will use vsce from the repo's 'node_modules' folder (which is a more scalable pattern for CI systems)

PR #55, opened 2018-04-10
Add lint (failing rules temporarily disabled)

StephenWeatherford:
#54

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,7 @@
 {
     "recommendations": [
         "eg2.tslint",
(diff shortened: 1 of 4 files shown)


.vscode/extensions.json
```diff
@@ -1,6 +1,7 @@
 {
     "recommendations": [
         "eg2.tslint",
+        "ms-vscode.azure-account",
```
EricJizbaMSFT: I don't think azure account is ever used for ARM extension?

StephenWeatherford: Oh, I guess you're right.

Issue #42, opened 2018-03-12
Fix and enable unit tests on Travis CI


EricJizbaMSFT:
There's a bunch of existing tests, but we're not running them with Travis CI. Unfortunately there's a lot of failing tests so we'll have to fix that before we can enable this in CI

StephenWeatherford:
Fixed through #56 and #61

PR #61, opened 2018-04-13
Fix remaining unit tests 

StephenWeatherford:
Suggest you review without whitespace: https://github.com/Microsoft/vscode-azurearmtools/pull/61/files?w=1

Disabled a couple suites. Most of the others were failing because the unit tests expected a certain list of functions in the completion lists, but new functions have been added since that time and it wasn't expecting those. Fixed by changing code to ignore new functions, otherwise unit tests will require multiple updates every time new functions are added.

### DIFF
--- a/test/PositionContext.test.ts
+++ b/test/PositionContext.test.ts
@@ -18,10 +18,6 @@ import { PositionContext } from "../src/PositionContext";
 
 import * as jsonTest from "./JSON.test";
 
-// TODO: Fix and re-enable
-function suiteTODODisabled(name: string, func: Function) {
-}
-
 suite("PositionContext", () => {
     suite("fromDocumentLineAndColumnIndexes(DeploymentTemplate,number,number)", () => {
         test("with null deploymentTemplate", () => {
@@ -1193,7 +1189,7 @@ suite("PositionContext", () => {
         }
     });
 
-    suiteTODODisabled("signatureHelp", () => {
+    suite("signatureHelp", () => {
         test("not in a TLE", () => {
             const dt = new DeploymentTemplate(`{ "a": "AA" }`, "id");
             const pc: PositionContext = dt.getContextFromDocumentCharacterIndex(`{ "a": "A`.length);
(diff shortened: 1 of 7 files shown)


StephenWeatherford:
Part of #42

test/PositionContext.test.ts
```diff
@@ -465,17 +461,41 @@ suite("PositionContext", () => {
...
+            // Make it easier to see missing names quickly
+            let actualNames = actualItems.map(item => item.name);
+            let expectedNames = expectedItems.map(item => typeof item === 'string' ? item : item.name);
+            assert.deepStrictEqual(actualNames, expectedNames, "Names in the completion items did not match");
+
+            assert.deepStrictEqual(actualItems, expectedItems);
+        }
+
+        // IMPORTANT
+        // We will only test against the completions in this list.  Any new functions added to the ExpressionMetadata.json file
```
EricJizbaMSFT: Is this really "too difficult" because people update "ExpressionMetadata.json" a lot or is it just because the tests weren't being run with CI and have gotten out of date?

If the file is actually updated frequently, then I'm okay with your change. But can we at least print the ones we're ignoring so that someone might fix them later? If there's no value in ever fixing them later - I'm not sure why we even have these tests in the first place

StephenWeatherford: They still add new functions, although less commonly than before (there are about twice as many as when these test were written).  Have you looked at the tests?  They are *very* specific on exactly what happens at every single index in every single string.  Which is overtesting.  I do want it to be easy to update the function metadata.  These tests make it hard.

The tests are very good at the general issue of handling Intellisense and are very valuable there. But that functionality should be tested separate of the exact function metadata, which is what I'm doing by restricting it to current functions.  Another possible approach would have been to copy the current data and place it into the test.

So there's no need to "fix" these tests later - this is not a kludge, I'm just restricting them to testing actual logic and not the metadata itself.  I don't think adding tests for the actual metadata would make a lot of sense unless we find things regressing.

StephenWeatherford: Added some comments to this effect.

PR #76, opened 2018-04-19
Fix some lint (7), part of #54

StephenWeatherford:


src/ParameterDefinition.ts
```diff
@@ -2,6 +2,8 @@
 // Copyright (c) Microsoft Corporation.  All rights reserved.
 // ----------------------------------------------------------------------------
 
+// tslint:disable:no-unused-expression
```
EricJizbaMSFT: nit: What's the reason for this? I like other cases where you put "// Grandfathered in" (assuming that's the reason)

src/Treeview.ts
```diff
@@ -277,13 +277,15 @@ export class JsonOutlineProvider implements vscode.TreeDataProvider<string> {
                     if (keyNode.properties[i].name._value.toUpperCase() === "type".toUpperCase()) {
                         let resourceType = keyNode.properties[i].value._value.toUpperCase();
                         icon = "resources.svg";
+                        // tslint:disable:no-unused-expression
```
EricJizbaMSFT: Urrgg there are so many cleaner ways to do this but I'll let you use the "Not my code" excuse if you want haha

StephenWeatherford: Yep. :-)

StephenWeatherford: Mostly didn't want to have to create templates with all those resource types to test.  Added grandfather clause.

PR #93, opened 2018-04-27
Use displayName tag and shorten name labels

### LINKED ISSUE #80: Json Outline pane doesn't use displayName tag
If a resource has tags ➝ displayName, Visual Studio's JSON outline view would display this value in the outline view rather than the 'name' variable. It seems this extension just uses the 'name' variable here. This is particularly useful to make the outline more readable when all the names come from variables/parameters.

Could this be supported for more friendly display names in the JSON outline pane?

StephenWeatherford:
Make labels in the tree more useful by:
1) using the displayName tag if it exists (like the VS extension)
2) shorten name labels (similar to the VS extension)
   e.g. "[(variables('a')+variables('abc'))]" => "(<a>+<abc>)"

Fixes #80

### DIFF
--- a/src/JSON.ts
+++ b/src/JSON.ts
@@ -498,6 +498,11 @@ export abstract class Value {
     }
 
     public abstract accept(visitor: Visitor): void;
+
+    /**
+     * A user-friendly string that represents this value, suitable for display in a label, etc.
+     */
+    public abstract toFriendlyString(): string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/JSON.ts
```diff
@@ -498,6 +498,11 @@ export abstract class Value {
     }
 
     public abstract accept(visitor: Visitor): void;
+
+    /**
+     * A user-friendly string that represents this value, suitable for display in a label, etc.
+     */
+    public abstract toFriendlyString(): string;
```
PrashanthCorp: If it is being used solely for labels, I'd name it `toLabel`. What friendly for labels may not be friendly for filename titles.

StephenWeatherford: I'm not sure it will be solely used for labels, thus etc.  I don't want to have something specific to just labels in the base classes, this makes it a bit more generic, comment specifies intent to help decide future direction when additional use cases are clear.

PR #110, opened 2018-05-16
Add tests to cover all known icons

StephenWeatherford:


StephenWeatherford:
ping

StephenWeatherford:
I'm more of a realist with unit tests.  This provides enough coverage that now I'm pretty confident in making changes in what used to be very risky code.  It's good enough for now and didn't take overly long to produce.  I also don't think changes to the structure are too likely but they could be fixed up then.

PR #111, opened 2018-05-16
Improve icon code

StephenWeatherford:


StephenWeatherford:
Passes the new tests

src/Treeview.ts
```diff
@@ -265,31 +292,26 @@ export class JsonOutlineProvider implements vscode.TreeDataProvider<string> {
         return JSON.stringify(result);
     }
 
-    // tslint:disable-next-line:cyclomatic-complexity // Grandfathered in
+    private getIcon(icons: [string, string][], itemName: string, defaultIcon: string) {
+        itemName = (itemName || "").toLowerCase();
+        let iconItem = icons.find(item => item[0].toLowerCase() === itemName);
+        return iconItem ? iconItem[1] : defaultIcon;
+    }
```
PrashanthCorp: Might want to annotate what the return type of `getIcon` is. I expect a file, but I'm not quite sure on reading the code what types `iconItem` and `icons` are. 
Moreover, the call sites below don't have the annotation, so I don't think I got the memo on `icon`'s type.

StephenWeatherford: I'll fix when adding that lint rule later, thanks.

PR #113, opened 2018-05-16
Fix some lint (11 - ordered imports)

StephenWeatherford:


StephenWeatherford:
#54

StephenWeatherford:
We already have fix-on-save enabled, but I didn't know about that, thanks.  Don't want it in the standard "lint" script because that's run on build integration and verification, added new script.

PR #116, opened 2018-05-19
Fix some lint (13)

StephenWeatherford:


src/AzureRMTools.ts
```diff
@@ -229,7 +231,7 @@ export class AzureRMTools {
 
                 this._diagnosticsCollection.set(document.uri, diagnostics);
             })
-            .catch((error: any) => {
```
EricJizbaMSFT: Mmm not a fan of changing these errors from 'any' to '{}'. Errors really could be 'any' (specifically I remember times when the error was a 'string' instead of an 'object').

That being said - is there an existing issue to clean up error/telemetry reporting? If so - I guess I don't really care what happens to this code since it'll hopefully change soon anyways

StephenWeatherford: Good point, I'm now suppressing them instead.  There are work items for cleaning this up but don't know when it will happen.

PR #136, opened 2018-10-17
Added new resource icons to treeview

rdbartram:
Added new SVGs to improve the visibility of resources in the JSON Outline


rdbartram:
just as an FYI, I would add many more icons, I just wanted to see if they would even get accepted.

EricJizbaMSFT:
@rdbartram thanks can you post screenshots in both light and dark theme?

rdbartram:
image


image

src/Treeview.ts
```diff
@@ -38,7 +38,13 @@ const resourceIcons: [string, string][] = [
     ["Microsoft.Compute/virtualMachines/extensions", "extensions.svg"],
     ["Microsoft.Network/networkSecurityGroups", "nsg.svg"],
     ["Microsoft.Network/networkInterfaces", "nic.svg"],
-    ["Microsoft.Network/publicIPAddresses", "publicip.svg"]
+    ["Microsoft.Network/publicIPAddresses", "publicip.svg"],
+    ["Microsoft.DocumentDB/databaseAccounts", "documentdb.svg"],
+    ["Microsoft.Web/serverfarms", "appserviceplan.svg"],
+    ["Microsoft.Sql/servers", "sqldatabase.svg"],
+    ["databases", "database.svg"],
```
EricJizbaMSFT: I don't see databses or firewallrules in your screenshot. Is this really the correct type? Every other type has "Microsoft/"

src/Treeview.ts
```diff
@@ -38,7 +38,13 @@ const resourceIcons: [string, string][] = [
     ["Microsoft.Compute/virtualMachines/extensions", "extensions.svg"],
     ["Microsoft.Network/networkSecurityGroups", "nsg.svg"],
     ["Microsoft.Network/networkInterfaces", "nic.svg"],
-    ["Microsoft.Network/publicIPAddresses", "publicip.svg"]
+    ["Microsoft.Network/publicIPAddresses", "publicip.svg"],
+    ["Microsoft.DocumentDB/databaseAccounts", "documentdb.svg"],
```
EricJizbaMSFT: The documentdb icon doesn't look great in dark theme. In the Cosmos DB extension we use this one: https://github.com/Microsoft/vscode-cosmosdb/blob/master/resources/icons/dark/CosmosDBAccount.svg

That being said - the current code only supports one icon per resource. Are you interested in refactoring to support themed icons? If not - I would prefer to wait on this particular icon until we fix that. If so - we can help you along the way. VS Code lets us specify the icon as either a string or an object with the path for both "light" and "dark". See here for more info: https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L6554

src/Treeview.ts
```diff
@@ -38,7 +38,13 @@ const resourceIcons: [string, string][] = [
...
+    ["Microsoft.Web/serverfarms", "appserviceplan.svg"],
```
EricJizbaMSFT: appserviceplan also doesn't look great in dark theme. I don't have an svg readily available, but this is what it looks like for me in the portal:
screen shot 2018-10-17 at 8 33 02 am

src/Treeview.ts
rdbartram: After you asked for the screenshots I looked up how to query for dark/light themes. I’ll take a look and make the appropriate changes.

src/Treeview.ts
rdbartram: The resources are listed like this in my arm template, so assume it’s right. It may be because they’re bested and inherit the namespace from the parent

EricJizbaMSFT: Okay I looked through the quick start templates repo and I'm seeing the same thing: https://github.com/Azure/azure-quickstart-templates

So sounds good to keep the type name as-is, but I'd still like to have screenshots of all icons you're adding

PR #154, opened 2019-01-07
Clean up a few things in the repo

EricJizbaMSFT:
1. Use 'opn' instead of 'open' to fix npm audit
1. Use typescript for gulp ([This PR](https://github.com/Microsoft/vscode-azurearmtools/pull/153) already switched us to gulp 4)
1. Make builds more consistent by always using Node 8
1. Commit the package-lock (as discussed/done in our other repos)


src/AzureRMTools.ts
```diff
@@ -12,8 +12,7 @@ import * as os from "os";
 import * as path from "path";
 import * as vscode from "vscode";
 
-// tslint:disable-next-line:no-var-requires
```
nturinski: How come we don't need to disable this tslint rule anymore?

PR #160, opened 2019-02-05
Make tests not dependent on production data

StephenWeatherford:


src/AzureRMAssets.ts
```diff
@@ -29,6 +29,14 @@ export class AzureRMAssets {
         return "azuresdk-3.0.0";
     }
 
+    // For test dependency injection only
+    public static setFunctionsMetadata(metadataString: string): void {
```
EricJizbaMSFT: I can't figure out where this is being used?

StephenWeatherford: I somehow lost a couple of files. Fixed.

PR #164, opened 2019-02-10
webpack

StephenWeatherford:
Also needed to convert to shared telemetry reporter.


src/AzureRMTools.ts
```diff
@@ -84,17 +84,15 @@ export class AzureRMTools {
     });
 
     constructor(context: vscode.ExtensionContext) {
-        ext.extensionContext = context;
+        ext.context = context;
```
EricJizbaMSFT: nit: Just remove this line. You do it in `activateInternal` already

StephenWeatherford:
> Would prefer to see a passing build before I approve, but otherwise looks good

Passing.

PR #176, opened 2019-02-27
Fix object disposed issue

### LINKED ISSUE #173: An error dialog pops up after closing one changed .json file
**Version:** 1.32.0-insider (user setup)
**Node.js**: 10.2.0
**OS:** All
**Extension Version**: [20190223.9](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=3530)

**Steps to reproduce:**
1. Open one .json file from VS Code.
2. Change the contents of the opened file -> Try to close the opened tab.
3. Click 'Save' or 'Don't Save' on the popped confirmation dialog -> Check the result.

**Expect Experiment:**
No error occurs.

**Actual Experiment**:
Pop up an error dialog with the below message.
image

StephenWeatherford:
Fixes #173 

The issue was that _diagnosticsCollection was disposed by disposing _deploymentTemplateFileSubscriptions, but was not set to null, so was still being accessed.

### DIFF
--- a/src/AzureRMTools.ts
+++ b/src/AzureRMTools.ts
@@ -170,7 +170,7 @@ export class AzureRMTools {
                     // are fired for all JSON files, not just deployment
                     // templates.
                     if (Object.keys(me._deploymentTemplates).length === 0) {
-                        me.initializeExtension(surveyInfo);
+                        me.hookDeploymentTemplateEvents(surveyInfo);
                     }
 
                     me._deploymentTemplates[documentUri] = deploymentTemplate;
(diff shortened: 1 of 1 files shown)


src/AzureRMTools.ts
```diff
@@ -418,8 +433,7 @@ export class AzureRMTools {
             // Code editor. The events will be registered again when the next
             // deployment template is opened.
             if (Object.keys(this._deploymentTemplates).length === 0 && this._deploymentTemplateFileSubscriptions) {
```
EricJizbaMSFT: nit: Since you're checking `this._deploymentTemplateFileSubscriptions` inside `unhookDeploymentFileEvents`, I would prefer to remove that check here

StephenWeatherford: I'll consider for later.

PR #183, opened 2019-03-05
Updating for new functions - newGuid, dateTime

sanjaiganesh:
Updating for new set of functions - newGuid, utcNow.

### DIFF
--- a/assets/ExpressionMetadata.json
+++ b/assets/ExpressionMetadata.json
@@ -100,6 +100,13 @@
             "maximumArguments": 1
         },
         {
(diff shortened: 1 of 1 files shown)


StephenWeatherford:
It will go out in the next release.

sanjaiganesh:
wow.. great turnaround.. Thanks @StephenWeatherford. Just FYI that these functions are live now and can be used with template deployment in the portal.

StephenWeatherford:
Good to know, thanks.  There's a bug fix I'd like to get out soon, so will probably release both soon.

PR #185, opened 2019-03-06
Reverting PR 183 (newguid and utcnow functions)

sanjaiganesh:
Reverting #183 to avoid false positives as their usage is scoped to parameter ->defaultValue


StephenWeatherford:
Will be fixing this: https://github.com/microsoft/vscode-azurearmtools/issues/243

PR #197, opened 2019-04-05
Add colorization of expressions

StephenWeatherford:
The current colorization here is designed so that it affects all .json and .jsonc files - therefore it has to be careful to only change the colorization of actual ARM template files. This is done by using injection to add rules to the built-in JSON/JSONC rules that detect the ARM schema and switch to our grammars if it's found (see detect-arm.injection.tmLanguage.json). This will go away when we switch to using our own language ID.

The change in tests from asdf -> asdg is simply because I use asdf as an easy keyword to search for UNDONEs while I code so I keep finding these. :-)

Known issues:
1) Currently while typing in a function call, you get red until you enter the parentheses. Working on changing that because it's annoying.
2) Having tested much with multi-line
3) Could use more testing while typing an entire expression


StephenWeatherford:
> I like the contrast between "[ and resourceGroup and wish variables had a similar amount of contrast with "[. Perhaps if "[ was always the same color as "Microsoft.Network/virtualNetworks"?

Unfortunately I don't have that level of control - I've only got so many scopes that are actually targeted by the current themes (and the themes vary in their contrast).  I hesitate to use the same color as "Microsoft.Network/virtualNetworks", because that's the normal string color, and Brian and I like that expression strings stand out as being different.  But I'll put this in the list to consider improvements after we've played around with it more.

Thanks!

PR #200, opened 2019-04-08
Update to UI 0.23.1

StephenWeatherford:


package.json
```diff
@@ -102,6 +102,6 @@
     "fs-extra": "^7.0.1",
     "moment": "^2.24.0",
     "opn": "^5.4.0",
-    "vscode-azureextensionui": "^0.20.4"
+    "vscode-azureextensionui": "^0.23.1"
```
EricJizbaMSFT: 0.21.0 bumped the minimum VS Code version to 1.26.0: https://github.com/Microsoft/vscode-azuretools/releases/tag/v0.21.0-ui

That was for wizard stuff which doesn't apply here, but still think this repo should bump the minimum engine

StephenWeatherford: Thanks!

PR #204, opened 2019-04-10
Remove unused `AzureRMToolsConfiguration`

EricJizbaMSFT:
I noticed a spike in telemetry for ARM and the 'loadConfiguration' event was the cause. I went to mark it as an activationEvent, but noticed `_azureRMToolsConfiguration` is never used. In fact, the README already says to use the more standard `telemetry.enableTelemetry` instead of `azurermtools.enableTelemetry`.

In addition, `_autoSave` is only used in `onDocumentSaved`, but I don't think that telemetry event has much value at all. It seems better _not_ to run code on every save just for telemetry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,13 +59,7 @@
     "configuration": {
       "type": "object",
       "title": "Azure Resource Manager Tools Configuration Settings",
-      "properties": {
-        "azurermtools.enableTelemetry": {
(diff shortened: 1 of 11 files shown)


EricJizbaMSFT:
Went ahead and turned on `noUnusedLocals` to prevent this from happening in the first place

PR #215, opened 2019-04-20
Saw/user ns vs func

StephenWeatherford:
Use same color but different scopes for user ns/func.  Will be no change in appearance for current themes, but using different scopes allows themes or users to target them separately.


grammars/arm-expression-string.tmLanguage.json
```diff
@@ -128,33 +129,37 @@
...
 				"1": {
-					"$comment": "namespace + user function name",
-					"name": "{{scope-usernamespaceandfunction}}"
+					"$comment": "user namespace (capturing group inside ns-userfunc)",
+					"name": "{{scope-usernamespace}}"
 				},
 				"2": {
-					"$comment": "dot from capturing group inside ns-userfunc",
+					"$comment": "dot (capturing group inside ns-userfunc)",
 					"name": "{{scope-usernamespaceandfunction-dot}}"
```
EricJizbaMSFT: Admittedly I don't fully understand the colorization stuff - but elsewhere you removed `usernamespaceandfunction`. Is this one still supposed to be here?

StephenWeatherford: Yeah, that's the dot between the namespace and function. Thanks for asking.

Issue #193, opened 2019-03-26
Random notifications with "documentColumnIndex cannot be greater than the line's maximum index"


jemag:
Repro steps:
Really hard to tell what causes this specific issue as it seems fairly random.

I think it might happen when autocomplete suggestions are triggered. I use a smaller vscode window where it only takes half of my screen. My only guess is that perhaps there are some situations where the autocompletion menu might not have enough space to render?!

Although this notification happens fairly often, I have not been able to determine the exact steps that will reproduce it 100% of the time.

Action: provideCompletionItems
Error type: ERR_ASSERTION
Error Message: documentColumnIndex (27) cannot be greater than the line's maximum index (26)

Version: 0.5.0
OS: Arch linux
VS Code Version: 1.32.2

<details>
<summary>Call Stack</summary>

```
Function.fromDocumentLineAndColumnIndexes (dist/extension.bundle.js:127:96193)
DeploymentTemplate.getContextFromDocumentLineAndColumnIndexes (dist/extension.bundle.js:127:92640)
Object.<anonymous> (dist/extension.bundle.js:127:125835)
Generator.next (<anonymous>)
module.exports.n (dist/extension.bundle.js:127:117380)
new Promise (<anonymous>)
module.exports.n (dist/extension.bundle.js:127:117157)
Object.<anonymous> (dist/extension.bundle.js:127:125744)
Object.<anonymous> (dist/extension.bundle.js:1:182883)
Generator.next (<anonymous>)
module.exports.n (dist/extension.bundle.js:1:181258)
new Promise (<anonymous>)
module.exports.n (dist/extension.bundle.js:1:181035)
Object.module.exports.t.callWithTelemetryAndErrorHandling (dist/extension.bundle.js:1:182800)
AzureRMTools.<anonymous> (dist/extension.bundle.js:127:125667)
Generator.next (<anonymous>)
module.exports.n (dist/extension.bundle.js:127:117380)
new Promise (<anonymous>)
module.exports.n (dist/extension.bundle.js:127:117157)
AzureRMTools.onProvideCompletionItems (dist/extension.bundle.js:127:125562)
Object.provideCompletionItems (dist/extension.bundle.js:127:121876)
define.provideCompletionItems.a.asPromise (extensionHostProcess.js:569:293)
t.asPromise.Promise (extensionHostProcess.js:48:58)
new Promise (<anonymous>)
Object.t.asPromise (extensionHostProcess.js:48:32)
[...]

PR #219, opened 2019-04-23
Suppress error display for UI interactions

StephenWeatherford:
Meant to stop the annoyance for people getting hit by https://github.com/Microsoft/vscode-azurearmtools/issues/193.  Still need to investigate properly.


nturinski:
Since you've discovered the cause of this issue in https://github.com/Microsoft/vscode-azurearmtools/issues/221, do you still want to suppress the errors?

StephenWeatherford:
> Since you've discovered the cause of this issue in #221, do you still want to suppress the errors?

Yes, I think it still makes sense. If we have an issue like this again, they'll get frequent error popups and we'll get telemetry anyway.

PR #223, opened 2019-04-24
Color/scope tweaks

StephenWeatherford:
1) Changed some categories (and therefore colors) for better contrast
1) Separate categories for parameters/variables
1) No separate category for runtime functions
1) Add separate category for logical functions
1) Color the string literal inside 'variables()' or 'parameters()' same as variables/parameters itself.
1) Use 'function' color for opening "[" and "]" of expressions (gives better contrast, plus this is about functions so it *kind* of makes sense.

Pre:
image

Post:
image


StephenWeatherford:
FYI, the PMs ended up deciding it was best as is.  We'll see how customers respond.

PR #242, opened 2019-05-04
Completions for deep parameter access

### LINKED ISSUE #124: Intellisense for parameter object .notation from defaultValue
At present there is no Intellisense completion for parameter objects in the same way that there is for variable objects or inbuilt function objects such as subscriptionId() or resourceGroup().

It would be great if it could add autocomplete based on the defaultValue for the objects if one is declared!

PS. Great extension!

StephenWeatherford:
Fixes #124 

This functionality already exists for variables that are of type object, this adds it for properties with a defaultValue of an object type.  

Example:
{
"$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
"contentVersion": "1.0.0.0",
"variables": {
"abc": {
"one": 1,
"two": {
"three": 2
}
},
"v1": "[variables('abc').two.three]",
"v2": "[parameters('def')]"
},
"parameters": {
"def": {
"type": "object",
"defaultValue": {
"one": 1,
"two": {
"three": 2
}
}
}
},
"resources": [],
"outputs": {
"a": {
"type": "string",
"value": "[concat(variables('v1'),variables('v2'))]"
}
}
image

### DIFF
--- a/src/AzureRMTools.ts
+++ b/src/AzureRMTools.ts
@@ -414,8 +414,8 @@ export class AzureRMTools {
                 this.suppressErrorDisplay = true;
 
                 const context: PositionContext = deploymentTemplate.getContextFromDocumentLineAndColumnIndexes(position.line, position.character);
-                if (context.completionItems) {
-                    let completionItemArray: Completion.Item[] = await context.completionItems;
+                if (context.getCompletionItems()) {
+                    let completionItemArray: Completion.Item[] = await context.getCompletionItems();
                     const completionItems: vscode.CompletionItem[] = [];
                     for (const completion of completionItemArray) {
                         const insertRange: vscode.Range = me.getVSCodeRangeFromSpan(deploymentTemplate, completion.insertSpan);
--- a/src/AzureRMTools.ts
+++ b/src/AzureRMTools.ts
@@ -414,42 +414,40 @@ export class AzureRMTools {
                 this.suppressErrorDisplay = true;
 
                 const context: PositionContext = deploymentTemplate.getContextFromDocumentLineAndColumnIndexes(position.line, position.character);
-                if (context.getCompletionItems()) {
-                    let completionItemArray: Completion.Item[] = await context.getCompletionItems();
-                    const completionItems: vscode.CompletionItem[] = [];
-                    for (const completion of completionItemArray) {
-                        const insertRange: vscode.Range = me.getVSCodeRangeFromSpan(deploymentTemplate, completion.insertSpan);
+                let completionItemArray: Completion.Item[] = await context.getCompletionItems();
+                const completionItems: vscode.CompletionItem[] = [];
+                for (const completion of completionItemArray) {
+                    const insertRange: vscode.Range = me.getVSCodeRangeFromSpan(deploymentTemplate, completion.insertSpan);
 
-                        const completionToAdd = new vscode.CompletionItem(completion.name);
-                        completionToAdd.range = insertRange;
-                        completionToAdd.insertText = new vscode.SnippetString(completion.insertText);
-                        completionToAdd.detail = completion.detail;
-                        completionToAdd.documentation = completion.description;
+                    const completionToAdd = new vscode.CompletionItem(completion.name);
+                    completionToAdd.range = insertRange;
+                    completionToAdd.insertText = new vscode.SnippetString(completion.insertText);
+                    completionToAdd.detail = completion.detail;
+                    completionToAdd.documentation = completion.description;
 
-                        switch (completion.kind) {
-                            case Completion.CompletionKind.Function:
-                                completionToAdd.kind = vscode.CompletionItemKind.Function;
-                                break;
+                    switch (completion.kind) {
+                        case Completion.CompletionKind.Function:
+                            completionToAdd.kind = vscode.CompletionItemKind.Function;
+                            break;
 
-                            case Completion.CompletionKind.Parameter:
-                            case Completion.CompletionKind.Variable:
-                                completionToAdd.kind = vscode.CompletionItemKind.Variable;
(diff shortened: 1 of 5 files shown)


src/AzureRMTools.ts
```diff
@@ -414,8 +414,8 @@ export class AzureRMTools {
                 this.suppressErrorDisplay = true;
 
                 const context: PositionContext = deploymentTemplate.getContextFromDocumentLineAndColumnIndexes(position.line, position.character);
-                if (context.completionItems) {
-                    let completionItemArray: Completion.Item[] = await context.completionItems;
+                if (context.getCompletionItems()) {
```
EricJizbaMSFT: I believe this should be:
```
let completionItemArray: Completion.Item[] = await context.getCompletionItems();
if (completionItemArray) {
```

Or just get rid of the `if` if `completionItemArray` is always defined. Right now you're starting the promise twice

src/PositionContext.ts
```diff
@@ -172,8 +172,8 @@ export class PositionContext {
         return this._hoverInfo;
     }
 
-    // tslint:disable-next-line:cyclomatic-complexity // Grandfathered in
-    public get completionItems(): Promise<Completion.Item[]> {
+    // tslint:disable-next-line:cyclomatic-complexity max-func-body-length // Grandfathered in
+    public async getCompletionItems(): Promise<Completion.Item[]> {
         if (this._completionItems === undefined) {
             this._completionItems = Promise.resolve([]);
```
EricJizbaMSFT: I'm not getting auto-completions for `resourceGroup()` anymore. I did a little digging and think it's related to my above comment and this line. You're basically calling `getCompletionItems` twice, which means the first sets this to an empty array here and the second just returns that empty array before the first call finishes getting the completions. I don't think there was a risk of race conditions before when this was synchronous, but I think now you have to be careful about that.

src/PositionContext.ts
```diff
@@ -237,54 +238,49 @@ export class PositionContext {
...
                         if (variableProperty) {
-
-                            const variableDefinition: Json.ObjectValue = Json.asObjectValue(variableProperty.value);
-                            if (variableDefinition) {
-
-                                const sourcePropertyDefinition: Json.ObjectValue = Json.asObjectValue(variableDefinition.getPropertyValueFromStack(sourcesNameStack));
+                            // If the variable's value is an object...
+                            const sourceVariableDefinition: Json.ObjectValue = Json.asObjectValue(variableProperty.value);
+                            if (sourceVariableDefinition) {
+                                this._completionItems = Promise.resolve(this.getDeepPropertyAccessCompletions(
```
EricJizbaMSFT: nit: If this function is async, you don't need `Promise.resolve` anymore

test/PositionContext.test.ts
```diff
@@ -1084,97 +1099,154 @@ suite("PositionContext", () => {
...
+
+                return {
+                    documentText: document,
+                    tokens
+                };
             }
+
+            suite("Parameter defaultValue deep completion for objects", () => {
+                let { documentText, tokens } = getDocumentAndMarkers({
+                    parameters: {
```
EricJizbaMSFT: So it's not obvious to me that I have to define a defaultValue to get completion for this. I also might want completion for properties not included in the defaultValue. Can I specify the schema for parameters of type object? I'm talking mainly about the "properties" object like [in this example](http://json-schema.org/learn/miscellaneous-examples.html):
```
{
  "$id": "https://example.com/person.schema.json",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
      "description": "The person's first name."
    },
    "lastName": {
      "type": "string",
      "description": "The person's last name."
    },
    "age": {
      "description": "Age in years which must be equal to or greater than zero.",
      "type": "integer",
      "minimum": 0
    }
  }
}
```

In other words, I would expect "firstName", "lastName", and "age" to be the completion entries instead of whatever is in a defaultValue.

src/PositionContext.ts
StephenWeatherford: Weird that the tests passed.

src/AzureRMTools.ts
StephenWeatherford: The problem with auto-rename, although don't know why the if was there in the first place.

src/PositionContext.ts
StephenWeatherford: Tests passed because they were unit tests, calling the get directly.  Added testing.

test/PositionContext.test.ts
StephenWeatherford: No, you can't currently do that in ARM templates, but good question.

test/PositionContext.test.ts
```diff
@@ -451,14 +451,37 @@ suite("PositionContext", () => {
...
+        function addCursor(documentText: string, markerIndex: number): string {
+            return `${documentText.slice(0, markerIndex)}<CURSOR>${documentText.slice(markerIndex)}`;
+        }
+
         function completionItemsTest(documentText: string, index: number, expectedCompletionItems: Completion.Item[]): void {
-            test(`with ${Utilities.escapeAndQuote(documentText)} at index ${index}`, () => {
+            const testName = `with ${Utilities.escapeAndQuote(addCursor(documentText, index))} at index ${index}`;
+            test(testName, async () => {
+                let keepInClosureForEasierDebugging = testName;
+                keepInClosureForEasierDebugging = keepInClosureForEasierDebugging;
```
EricJizbaMSFT: Intentional to check in this and below commented-out code?

StephenWeatherford: 1 yes, 2 no

PR #246, opened 2019-05-14
extend microsoft contrib

StephenWeatherford:


tslint.json
```diff
@@ -1,79 +1,33 @@
 {
     "extends": "tslint-microsoft-contrib",
```
EricJizbaMSFT: Not really sure what this PR does since it looks like it already extended `tslint-microsoft-contrib`, but looks fine

StephenWeatherford: Sorry, I was originally looking at the wrong project when I originally made the branch. This just cleans some things up and gets closer to lint-free.

PR #247, opened 2019-05-18
Lint-free

### LINKED ISSUE #54: Become lint free

StephenWeatherford:
Fixes #54

### DIFF
--- a/src/AzureRMAssets.ts
+++ b/src/AzureRMAssets.ts
@@ -15,6 +15,7 @@ export class AzureRMAssets {
 
     // For test dependency injection only
     public static setFunctionsMetadata(metadataString: string): void {
+        // tslint:disable-next-line:typedef
         AzureRMAssets._functionsMetadataPromise = new Promise<FunctionsMetadata>(async (resolve, reject) => {
             let array = FunctionMetadata.fromString(metadataString);
             resolve(new FunctionsMetadata(array));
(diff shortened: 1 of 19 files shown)


tslint.json
```diff
@@ -25,9 +25,12 @@
...
         "no-backbone-get-set-outside-model": false,
         "no-implicit-dependencies": [
-            false, // TODO
+            true,
             [
-                "vscode"
+                "vscode",
+                "gulp",
+                "vscode-azureextensiondev",
+                "mocha"
```
EricJizbaMSFT: Suggest adding a `tslint.json` file to your test folder that allows all "dev" dependencies

Something like this: https://github.com/microsoft/vscode-azurefunctions/blob/master/test/tslint.json#L5

StephenWeatherford: Cool, thanks.

src/AzureRMAssets.ts
```diff
@@ -15,6 +15,7 @@ export class AzureRMAssets {
 
     // For test dependency injection only
     public static setFunctionsMetadata(metadataString: string): void {
+        // tslint:disable-next-line:typedef
```
nturinski: Why disable this rule rather than type defining the property??

StephenWeatherford: Because I hate providing obvious stuff that the compiler is already smart enough to infer anyway.

Issue #250, opened 2019-05-30
Does not correctly handle colorization when a string starts with a bracket but does not end with a bracket


StephenWeatherford:
- [x] fix colorization
- [ ] fix validation

See https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates#syntax:
> To have a literal string start with a bracket [, but not have it interpreted as an expression, add an extra bracket to start the string with [[.

But it turns out you only need the [[ if your string starts with a bracket *and ends with a bracket*.  I.e.:
  
   "value": "[ChefInSpec]InstalledApplicationLinuxResource1;AttributesYmlContent"

is perfectly valid and is interpreted as a string, since it does not end with a bracket.

StephenWeatherford:
Example template:

```arm-deployment
{
  "$schema": "hts://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "resources": [
  ],
  "outputs": {
    "one": {
      "type": "string",
      "value": "one"
    },
    "two": {
      "type": "string",
      "value": "[[two]" // This is the string value '[two]'
    },
    "three": {
      "type": "string",
      "value": "[two]andthree" // this is the string value '[two]andthree'
    }
    //"four": {
    //  "type": "string",
    //  "value": "[four]" // this is invalid
    //}
  }
}
```

Output:
```
DeploymentName          : azuredeploy5
ResourceGroupName       : deleteme
ProvisioningState       : Succeeded
Timestamp               : 5/30/19 11:27:23 PM
Mode                    : Incremental
TemplateLink            :
Parameters              :
Outputs                 :
                          Name             Type                       Value
                          ===============  =========================  ==========
                          one              String                     one
                          two              String                     [two]
                          three            String                     [two]andthree
```

StephenWeatherford:
Mistakingly closed

StephenWeatherford:
Another example template:
```
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "outputs": {
        // 1: Starts with [ but doesn't end with ] -> it is an unchanged string
        "output1a": {
            "type": "string",
            "value": "[one]two" // -> string: "[one]two" 
        },
        "output1b": {
            "type": "string",
            "value": "[one]two" // -> string: "[one]two"
        },
        "output1c": {
[...]

PR #252, opened 2019-06-05
Fixed issue #250 - handles case where string starts with bracket but does not end with bracket

sbuttar:
Added test cases relevant to issue
Added logic to TLE.ts to handle case where string starts with bracket but does not end with bracket
Modified test case in DeploymentTemplate.test.ts

### DIFF
--- a/src/TLE.ts
+++ b/src/TLE.ts
@@ -880,34 +880,15 @@ export class Parser {
         const errors: language.Issue[] = [];
         let result: { stringValue: string; leftSquareBracketToken: Token; expression: Value; rightSquareBracketToken: Token; errors: language.Issue[] } = { stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors };
 
-        if (3 <= stringValue.length && stringValue.substr(1, 2) === "[[") {
+        if (
+            stringValue[1] !== "[" // Does not start with '['
+            || stringValue.substr(1, 2) === "[[" // Starts with '[['
+            || (stringValue[1] === '[' && stringValue[stringValue.length - 1] === ']') // Starts with '[' but doesn't end with ']'
+        ) {
+            // This is considered a plain string
             result.expression = new StringValue(Token.createQuotedString(0, stringValue));
-        } else if (stringValue[1] === '[' && stringValue[stringValue.length - 2] === ']') {
-            if (3 <= stringValue.length && stringValue.substr(1, 2) === "[[") {
-                expression = new StringValue(Token.createQuotedString(0, stringValue));
-            } else {
-                result = Parser.expressionValidation(stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors);
-            }
         } else {
-            let stringCase = false;
-            let index = 0;
-            if (stringValue[1] === '[') {
-                for (let rightBracket of stringValue) {
-                    ++index;
-                    if (rightBracket === ']') {
-                        for (index; index < stringValue.length; ++index) {
-                            if (stringValue[index] === '"') {
-                                result.expression = new StringValue(Token.createQuotedString(0, stringValue));
-                                stringCase = true;
-                                break;
-                            }
-                        }
-                    }
-                }
-            }
-            if (stringCase === false) {
-                result = Parser.expressionValidation(stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors);
-            }
+            result = Parser.expressionValidation(stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors);
         }
         return new ParseResult(result.leftSquareBracketToken, result.expression, result.rightSquareBracketToken, result.errors);
     }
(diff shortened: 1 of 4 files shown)


test/TLE.test.ts
```diff
@@ -581,6 +581,28 @@ suite("TLE", () => {
                 assert.deepStrictEqual([], pr.errors);
             });
 
+            test("with two right square brackets, literal, and closing left square bracket", () => {
```
StephenWeatherford: left/right transposed in comment

sbuttar: Fixed!

src/TLE.ts
```diff
@@ -874,65 +874,95 @@ export class Parser {
...
-        let leftSquareBracketToken: Token = null;
+        const leftSquareBracketToken: Token = null;
         let expression: Value = null;
-        let rightSquareBracketToken: Token = null;
-        let errors: language.Issue[] = [];
+        const rightSquareBracketToken: Token = null;
+        const errors: language.Issue[] = [];
+        let result: { stringValue: string; leftSquareBracketToken: Token; expression: Value; rightSquareBracketToken: Token; errors: language.Issue[] } = { stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors };
 
         if (3 <= stringValue.length && stringValue.substr(1, 2) === "[[") {
```
StephenWeatherford: These changes seem more complex than necessary.  Also, note that this code was ported from C# - the checks against the string length aren't necessary, out of bounds string indices just return undefined.

I'd think the following single change would do it:

        if (
            stringValue[1] !== "[" // Does not start with '['
            || stringValue.substr(1, 2) === "[[" // Starts with '[['
            || (stringValue[1] === '[' && stringValue[stringValue.length - 1] === ']') // Starts with '[' but doesn't end with ']'
        ) {
            // This is considered a plain string
            expression = new StringValue(Token.createQuotedString(0, stringValue));
        } else {
          // This is considered an expression

With that first check I added to verify it starts with '[', you should also be able to get rid of the following if/else (this will keep us from tokenizing the entire string when it doesn't even start with '[' - I didn't write this code :-)).


            if (!tokenizer.hasCurrent() || tokenizer.current.getType() !== TokenType.LeftSquareBracket) {
                expression = new StringValue(Token.createQuotedString(0, stringValue));
            } else {

StephenWeatherford: Note: This exposes some incorrect testcases - currently since the string is tokenized before checking for [, whitespace is thrown away, and it considers strings like " [expr()] " to be an expression, but when I run that through the portal, it considers it a string, not an expression.

src/TLE.ts
```diff
@@ -874,65 +874,78 @@ export class Parser {
...
+        const leftSquareBracketToken: Token = null;
         let expression: Value = null;
-        let rightSquareBracketToken: Token = null;
-        let errors: language.Issue[] = [];
+        const rightSquareBracketToken: Token = null;
+        const errors: language.Issue[] = [];
+        let result: { stringValue: string; leftSquareBracketToken: Token; expression: Value; rightSquareBracketToken: Token; errors: language.Issue[] } = { stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors };
+
+        let closingBracketLocation = stringValue.length - 1;
+        for (closingBracketLocation; closingBracketLocation > 0; closingBracketLocation--) {
```
StephenWeatherford: What's the intention of this code?  Is it to handle the case where the string is unterminated?  Again, getting too complicated.

Let's get rid of this and turn the condition around:

    if (stringValue.startsWith('"[') && stringValue.endsWith(']"') && !stringValue.startsWith('"[[')) {
       // This is considered an expression
…



            expression = new StringValue(Token.createQuotedString(0, stringValue)); 


            stringValue[1] !== "[" // Does not start with '[' 




            || stringValue.substr(1, 2) === "[[" // Starts with '[[' 




            || (stringValue[1] === '[' && stringValue[closingBracketLocation] !== ']' && stringValue.indexOf(']') >= 0) // Starts with '[' but doesn't end with ']' 




        ) { 




            // This is considered a plain string

test/TLE.test.ts
```diff
@@ -581,6 +581,28 @@ suite("TLE", () => {
...
+                assert.notEqual(null, pr);
+                assert.equal(null, pr.leftSquareBracketToken);
+                assert.deepStrictEqual(
+                    new TLE.StringValue(TLE.Token.createQuotedString(0, "\"[two]andthree\"")),
+                    pr.expression);
+                assert.equal(null, pr.rightSquareBracketToken);
+                assert.deepStrictEqual([], pr.errors);
+            });
+
             test("with function name without parentheses, arguments, or right square bracket", () => {
```
StephenWeatherford: Are there testcases handling an unterminated string?  If not, please add once, e.g.:

{
    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "variables": {
        "nsgId": "not terminated

src/TLE.ts
```diff
@@ -874,65 +874,78 @@ export class Parser {
...
+        }
+        return new ParseResult(result.leftSquareBracketToken, result.expression, result.rightSquareBracketToken, result.errors);
+    }
 
-            if (!tokenizer.hasCurrent() || tokenizer.current.getType() !== TokenType.LeftSquareBracket) {
-                expression = new StringValue(Token.createQuotedString(0, stringValue));
-            } else {
-                leftSquareBracketToken = tokenizer.current;
-                tokenizer.next();
+    private static expressionValidation(stringValue: string, leftSquareBracketToken: Token, expression: Value, rightSquareBracketToken: Token, errors: language.Issue[]): { stringValue: string; leftSquareBracketToken: Token; expression: Value; rightSquareBracketToken: Token; errors: language.Issue[] } {
```
StephenWeatherford: method names should be verbs, suggest "parseStringAsExpression"

src/TLE.ts
```diff
@@ -874,65 +874,95 @@ export class Parser {
...
         if (3 <= stringValue.length && stringValue.substr(1, 2) === "[[") {
+            result.expression = new StringValue(Token.createQuotedString(0, stringValue));
+        } else if (stringValue[1] === '[' && stringValue[stringValue.length - 2] === ']') {
+            if (3 <= stringValue.length && stringValue.substr(1, 2) === "[[") {
+                expression = new StringValue(Token.createQuotedString(0, stringValue));
+            } else {
+                result = Parser.parseStringAsExpression(stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors);
+            }
+        } else {
+            let stringCase = false;
```
StephenWeatherford: Why is this new code necessary?

src/TLE.ts
StephenWeatherford: should be able to get rid of the "3 <= stringValue.length" and other string length checks

src/TLE.ts
```diff
@@ -874,65 +874,95 @@ export class Parser {
...
+                }
+            }
+            if (stringCase === false) {
+                result = Parser.parseStringAsExpression(stringValue, leftSquareBracketToken, expression, rightSquareBracketToken, errors);
+            }
+        }
+        return new ParseResult(result.leftSquareBracketToken, result.expression, result.rightSquareBracketToken, result.errors);
+    }
+
+    private static parseStringAsExpression(stringValue: string, leftSquareBracketToken: Token, expression: Value, rightSquareBracketToken: Token, errors: language.Issue[]): { stringValue: string; leftSquareBracketToken: Token; expression: Value; rightSquareBracketToken: Token; errors: language.Issue[] } {
```
StephenWeatherford: I'm not convinced you've gained anything by refactoring out praseStringAsExpression.  It seems like it's only made it more complex and harder to read, especially with the way you're passing parameters in and out.

I really don't think all the changes you have are necessary.  I would start again with the original function for this file and make just the changes necessary.

StephenWeatherford: Again look at my suggested if/then.  I think it really should be as easy as getting rid of the current "if (3 <= strintgValue.length…" statement and replacing it with my suggestion:

   if (
        stringValue[1] !== "[" // Does not start with '['
        || stringValue.substr(1, 2) === "[[" // Starts with '[['
        || (stringValue[1] === '[' && stringValue[stringValue.length - 1] === ']') // Starts with '[' but doesn't end with ']'
    ) {
        // This is considered a plain string
        expression = new StringValue(Token.createQuotedString(0, stringValue));
    } else {
      // This is considered an expression
      let tokenizer = Tokenizer.fromString(stringValue);
      tokenizer.next();
      ...

PR #257, opened 2019-06-18
Change publisher from "msazurermtools" to "ms-azuretools"

EricJizbaMSFT:


StephenWeatherford:
We were discussing whether we wanted to rename this extension when it gets the new language server, or maybe have a way to preview it for a limited audience first, so this might be a convenient way of accomplishing this, so please wait before publishing with the new name.  Thanks!

EricJizbaMSFT:
@StephenWeatherford please see offline email. @fiveisprime wants us to change this ASAP because the marketplace already updated it for us https://marketplace.visualstudio.com/items?itemName=ms-azuretools.azurerm-vscode-tools

StephenWeatherford:
Okay

PR #259, opened 2019-07-17
Fix colorization for #250, strings vs expressions boundary cases

### LINKED ISSUE #250: Does not correctly handle colorization when a string starts with a bracket but does not end with a bracket

StephenWeatherford:
1) Allow comments in grammar preprocess section by ignoring keys beginning with "$", also a little clean-up
2) Correctly express some additional rules: a string must begin with [ and end with ] to be considered an expression.  Thus something like "[pre]post" is not an expression and not an error, it's just a regular JSON string
3) Given this, a few of the existing testcases are no longer valid and were removed

Fixes the colorization portion of #250


StephenWeatherford:
@sbuttar

PR #262, opened 2019-07-25
Bump to latest shared packages

EricJizbaMSFT:
There were a few breaking versions to the UI package in this, but this one related to `IActionContext` was the only one that affected this repo: https://github.com/microsoft/vscode-azuretools/releases/tag/v0.25.0-ui


EricJizbaMSFT:
ping @nturinski @StephenWeatherford @PrashanthCorp

src/AzureRMTools.ts
```diff
@@ -137,7 +133,7 @@ export class AzureRMTools {
                 // If the documentUri is not in our dictionary of deployment templates, then we
                 // know that this document was opened (as opposed to changed/updated).
                 let stopwatch: Stopwatch;
-                if (!me._deploymentTemplates[documentUri]) {
+                if (!this._deploymentTemplates[documentUri]) {
```
PrashanthCorp: Don't these changes seem orthogonal to bumping the shared packages?

EricJizbaMSFT: I _have_ to change from `(this: IActionContext)` to `(actionContext: IActionContext)` because of a breaking change in the UI package (the build won't even pass if I don't do that). I suppose I could've left the `me` variable as-is, but the only reason we did that in the first place was because actionContext was `this`. Now that it's not `this`, it feels directly related to clean that up.

PR #274, opened 2019-08-21
Clarify and simplify ARM term

StephenWeatherford:


package.json
```diff
@@ -58,7 +58,7 @@
     ],
     "configuration": {
       "type": "object",
-      "title": "Azure Resource Manager Tools Configuration Settings",
+      "title": "ARM Tools Configuration Settings",
```
alex-frankel: Where does this string get displayed?

StephenWeatherford: In the settings dialog:

image

StephenWeatherford:
Ping.  I still need approval from a dev on aztools.

README.dev.md
```diff
@@ -1,5 +1,5 @@
-# Azure Resource Manager Tools
-A JSON language extension to support Azure Resource Manager deployment templates and template language expressions.
+# Azure Resource Manager ("ARM") Tools
```
EricJizbaMSFT: Yeah I don't think you want me to review this 😂 I can't find any other documentation that does the `("ARM")` thing and it looks a little ugly to me...

None of these have it:
https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-overview
https://azure.microsoft.com/en-us/features/resource-manager/
https://docs.microsoft.com/en-us/azure/azure-resource-manager/

src/AzureRMTools.ts
```diff
@@ -28,7 +28,7 @@ import { UnrecognizedFunctionIssue } from "./UnrecognizedFunctionIssue";
 export async function activateInternal(context: vscode.ExtensionContext, perfStats: { loadStartTime: number; loadEndTime: number }): Promise<void> {
     ext.context = context;
     ext.reporter = createTelemetryReporter(context);
-    ext.outputChannel = vscode.window.createOutputChannel("Azure Resource Manager Tools");
+    ext.outputChannel = vscode.window.createOutputChannel("ARM Tools");
```
PrashanthCorp: I feel @EricJizbaMSFT has a valid point: `ARM` doesn't show up any azure related pages on Bing or Google search, but Azure ARM does. `Azure Resource Manager Tools` seems more explanatory. If it's too long a name, `Azure Tools - ARM` might be less ambiguous.

Q: Is `ARM` a branding Azure uses externally? I've seen MSFT employees use it, and as abbreviations inside Azure Resource Manager pages. Maybe @bmoore-msft can answer this.

StephenWeatherford: This came from legal, see forwarded e-mail.

package.json
EricJizbaMSFT: Okay lol but `Tools Configuration Settings` is clearly the problem here - not `Azure Resource Manager`

src/AzureRMTools.ts
EricJizbaMSFT: The email still says "ARM" is discouraged. Why can't we just use `Azure Resource Manager` everywhere and follow recommended best practices? It's not _THAT_ long. Also we don't have `Tools` in our other extension names, which will help shorten it

package.json
StephenWeatherford: True on that one, looks like it actually never gets displayed.

src/AzureRMTools.ts
StephenWeatherford: It gets rather unwieldy in some places, like the language name "Azure Resource Manager Deployment Template", "Azure Resource Manager Language Server", and as the source for diagnostic issues.

StephenWeatherford: The point about "tools" is good.  I guess we could just call it "Azure Resource Manager".  Maybe "Azure Resource Manager Templates"??

bmoore-msft: "ARM" is in formal - in tooling/docs we should be using "Azure Resource Manager".  In GitHub PR comments we can use ARM all we want ;)

StephenWeatherford:
Superseded by #290

PR #278, opened 2019-08-23
.net acquisition

StephenWeatherford:
Acquire .dotnet runtime at extension activation (because bundling with the extension would require us to bundle all platform versions)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,13 +32,14 @@
   "icon": "AzureRMTools128x128.png",
   "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
   "engines": {
-    "vscode": "^1.28.0"
+    "vscode": "^1.31.0"
   },
   "preview": true,
   "activationEvents": [
     "onLanguage:arm-deployment",
     "onLanguage:json",
-    "onLanguage:jsonc"
+    "onLanguage:jsonc",
+    "onCommand:azurerm-vscode-tools.uninstallDotnet"
   ],
   "contributes": {
     "grammars": [
@@ -106,6 +107,13 @@
           ".arm"
         ]
       }
+    ],
+    "commands": [
+      {
+        "category": "ARM Tools",
+        "title": "Uninstall dotnet core for this extension",
+        "command": "azurerm-vscode-tools.uninstallDotnet"
+      }
     ]
   },
   "main": "main",
@@ -131,8 +139,10 @@
   "devDependencies": {
     "@types/fs-extra": "^5.0.5",
     "@types/mocha": "^2.2.41",
-    "@types/node": "^8.10.48",
+    "@types/node": "^12.0.0",
     "@types/opn": "^5.5.0",
+    "@types/rimraf": "2.0.2",
+    "@types/shelljs": "0.8.3",
     "gulp": "^4.0.1",
     "mocha": "^5.2.0",
     "mocha-junit-reporter": "^1.22.0",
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -106,6 +107,13 @@
           ".arm"
         ]
       }
+    ],
+    "commands": [
```
NTaylorMullen: Hmm, when would you want users to utilize this command? Seems odd to expose it as part of the extension for others to utilize.

src/AzureRMTools.ts
```diff
@@ -209,20 +213,68 @@ export class AzureRMTools {
...
 
-            if (ext.addCompletionDiagnostic) {
-                // Add a diagnostic to indicate expression validation is done (for testing)
-                diagnostics.push(<vscode.Diagnostic>{
-                    severity: vscode.DiagnosticSeverity.Information,
-                    message: expressionsDiagnosticsCompletionMessage,
-                    source: expressionsDiagnosticsSource,
-                    code: ""
-                });
+            let languageServerLanguageDiagnostic = this.getLanguageServerStateDiagnostic();
```
NTaylorMullen: `const` here and several other places in the PR.

src/AzureRMTools.ts
```diff
@@ -209,20 +213,68 @@ export class AzureRMTools {
...
+            if (languageServerLanguageDiagnostic) {
+                diagnostics.push(languageServerLanguageDiagnostic);
+            }
+
+            let completionDiagnostic = this.getCompletionDiagnostic();
+            if (completionDiagnostic) {
+                diagnostics.push(completionDiagnostic);
             }
 
             this._diagnosticsCollection.set(document.uri, diagnostics);
```
NTaylorMullen: Curious: Do you guys maintain your own diagnostics collection? The `LanguageServerClient` can handle this transparently and provide access to the underlying diagnostics collection

src/languageclient/startArmLanguageServer.ts
```diff
@@ -1,53 +1,54 @@
...
 
-export function startArmLanguageServer(context: ExtensionContext): void {
-    callWithTelemetryAndErrorHandlingSync('startArmLanguageClient', () => {
+export async function startArmLanguageServer(): Promise<void> {
+    languageServerState = LanguageServerState.Starting;
+    try {
         // The server is implemented in .NET Core. We run it by calling 'dotnet' with the dll as an argument
+        let serverDllPath: string = findLanguageServer(ext.context);
+        let dotnetExePath: string = await acquireDotnet(serverDllPath);
+        await ensureDependencies(dotnetExePath, serverDllPath);
```
NTaylorMullen: So note: This will attempt to boot your server with 0 arguments. If there's a piece of your server that you want to try out (any complex dependencies, doubt you have native ones) I'd be sure that call path is hit during this boot to ensure that those assemblies can be loaded/executed properly.

package.json
StephenWeatherford: Frankly it's there mostly for testing purposes.  I don't see that it causes any harm.

src/AzureRMTools.ts
StephenWeatherford: Personally I'm not a fan of that rule, it's annoying when you can't change values during debugging sessions.  If I reconsider I'll turn on the lint rule.

src/languageclient/startArmLanguageServer.ts
StephenWeatherford: Hmm, good point.  I don't think we ever actually hit needing additional dependencies, though. I haven't seen it run yet.

I guess I'd need to add a hidden command-line argument to somehow hit the code analysis logic and then quit, since the server never exits normally.

src/AzureRMTools.ts
StephenWeatherford: Yes, we do.  I did notice that recently, might look into merging them, although there appears to be no harm in having two separate ones. Thx.

src/languageclient/startArmLanguageServer.ts
StephenWeatherford: I'm not actually clear on what dependencies can be missing, do you know?

StephenWeatherford: Thanks!

PR #285, opened 2019-09-10
Always validate file if langid=ARM

### LINKED ISSUE #284: Expression diagnostics don't show up if the schema is wrong but the user switches manually to ARM Language Server
Functionality should be enabled based only on whether langid=arm-deployment (though we can automatically switch to that langid if the schema matches).

StephenWeatherford:
Fixes #284

### DIFF
--- a/src/AzureRMTools.ts
+++ b/src/AzureRMTools.ts
@@ -156,31 +156,13 @@ export class AzureRMTools {
 
                         // Make sure the language ID is set to ARM deployment template
                         if (document.languageId !== armDeploymentLanguageId) {
-                            vscode.languages.setTextDocumentLanguage(document, armDeploymentLanguageId);
-
                             // The document will be reloaded, firing this event again with the new langid
-                            actionContext.telemetry.properties.switchedToArm = 'true';
-                            actionContext.telemetry.properties.docLangId = document.languageId;
-                            actionContext.telemetry.properties.docExtension = path.extname(document.fileName);
-                            actionContext.telemetry.suppressIfSuccessful = false;
+                            AzureRMTools.setLanguageToArm(document, actionContext);
                             return;
                         }
 
-                        ext.reporter.sendTelemetryEvent(
-                            "Deployment Template Opened",
-                            {
-                                docLangId: document.languageId,
-                                docExtension: path.extname(document.fileName),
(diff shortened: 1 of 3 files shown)


StephenWeatherford:
Ping

src/AzureRMTools.ts
```diff
@@ -120,72 +120,56 @@ export class AzureRMTools {
...
-                    // If the documentUri is not in our dictionary of deployment templates, then we
-                    // know that this document was just opened (as opposed to changed/updated).
-                    let stopwatch: Stopwatch;
-                    const documentUri: string = document.uri.toString();
-                    if (!this._deploymentTemplates[documentUri]) {
-                        stopwatch = new Stopwatch();
-                        stopwatch.start();
-                    }
+            const stopwatch = new Stopwatch();
+            stopwatch.start();
```
PrashanthCorp: Does this need to be accompanied by a stop?

src/AzureRMTools.ts
```diff
@@ -120,72 +120,56 @@ export class AzureRMTools {
...
-                                    varsCount: deploymentTemplate.variableDefinitions.length,
-                                    namespacesCount: deploymentTemplate.namespaceDefinitions.length
-                                });
-
-                            this.logFunctionCounts(deploymentTemplate);
+                    if (isNewlyOpened) {
+                        // A deployment template has been opened (as opposed to having been tabbed to)
+
+                        // Make sure the language ID is set to ARM deployment template
+                        if (document.languageId !== armDeploymentLanguageId) {
```
PrashanthCorp: What are the other possible values it can hold? I can see it having JSON. 
I'm wondering if modifying this might affect another extension that's also looking into this file.

src/AzureRMTools.ts
StephenWeatherford: No, it's just a thin wrapper on Date.now(), but thanks for asking.

src/AzureRMTools.ts
StephenWeatherford: We've notified the extensions we're aware of that might be affected. Ultimately the user has control of this, and can turn it off our automatic behavior with a setting.  Again, good question.

This was the solution suggested by vscode guys.

PR #299, opened 2019-09-20
Refactor PositionContext (post-release)

StephenWeatherford:
This PR is easiest to follow if you go by each commit sequentially.  No functionality changes.

1) Extracted functions from getCompletionItems
2) Add CachedValue/CachedPromise for better readability and easier later transformation to strict null checking
3) Convert to fully-async code

PR #300, opened 2019-09-21
Enable strict null checking, part 1

StephenWeatherford:
Depends on https://github.com/microsoft/vscode-azurearmtools/pull/299

Moving towards strict null enabled

### DIFF
--- a/src/TLE.ts
+++ b/src/TLE.ts
@@ -794,9 +794,10 @@ export class UndefinedVariablePropertyVisitor extends Visitor {
     }
 
     private addIssue(tlePropertyAccess: PropertyAccess): void {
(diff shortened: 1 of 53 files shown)


src/TLE.ts
```diff
@@ -788,10 +794,11 @@ export class UndefinedVariablePropertyVisitor extends Visitor {
     }
 
     private addIssue(tlePropertyAccess: PropertyAccess): void {
-        const propertyName: string = tlePropertyAccess.nameToken.stringValue;
+        const propertyName: string = tlePropertyAccess.nameToken ? tlePropertyAccess.nameToken.stringValue : "(unknown)";
```
bhsubra: nit: Can we put tlePropertyAccess.nameToken in a local since it's used in multiple places.

PR #303, opened 2019-09-24
Show and telemetrate lang server nuget version

StephenWeatherford:


bhsubra:
Changes look good to me.

PR #309, opened 2019-09-28
Improve description for uniqueString

StephenWeatherford:
After reading the existing description, I had to look it up to understand how to use.

image
image

### DIFF
--- a/assets/ExpressionMetadata.json
+++ b/assets/ExpressionMetadata.json
@@ -623,8 +623,8 @@
         },
         {
             "name": "uniqueString",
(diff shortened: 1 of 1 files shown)


StephenWeatherford:
I'm thinking of adding more info like this (probably not as much for most) for others as well.

bmoore-msft:
@mumian - should be able to help...

StephenWeatherford:
@bmoore-msft This comes straight from the docs, so the text doesn't really need reviewing, just more whether you like the longer format.

StephenWeatherford:
ping

StephenWeatherford:
Thanks!

PR #321, opened 2019-10-08
Add static support for snippets

StephenWeatherford:
Taking over the snippet functionality from https://github.com/sam-cogan/arm-snippets-vscode.  This first version is just a static set of snippets.  For JSON/JSONC, it contains only snippets for creating the skeleton of an ARM template or ARM parameters file.  For arm-template, it contains all the current snippets from https://github.com/sam-cogan/arm-snippets-vscode.

Need a write-up for the readme.


assets/armsnippets.jsonc
```diff
@@ -0,0 +1,2406 @@
+{ /*
```
bhsubra: Can we fix the formatting here?

assets/jsonsnippets.jsonc
```diff
@@ -0,0 +1,31 @@
+{
+    /*
```
bhsubra: Can we fix the formatting here?

StephenWeatherford: There are some issues I'm going to need to go through.  I consider this just a initial checkin.

StephenWeatherford: I'd fix them now but I'd rather have an initial checkin that has exactly the previous snippets so we can track changes.

PR #338, opened 2019-10-22
Various snippet fixes and add snippet tests

StephenWeatherford:
1) Tabs to spaces
2) Added missing commas
3) Removed trailing commas (before some had them, some didn't)
4) Remove Property-Value, because it's for .params files
5) Consistently use single quotes around field names that are inside expressions
6) Various fixes so there are no syntax errors
7) Add tests that insert and validate each snippet (currently not checking for schema errors, just syntax and template validation)


assets/armsnippets.jsonc
```diff
@@ -54,15 +54,6 @@
         ],
         "description": "ARM Template output"
     },
-    "Parameter Value": {
```
alex-frankel: is there a reason we got rid of this? Is this a snippet to help me add a parameter?

StephenWeatherford: It turned out it was a snippet to add a parameter to a template *params* file, which doesn't make sense for arm-template.  However, it does make sense for json, so I'll add it there.

PR #360, opened 2019-10-28
Add some requested telemetry

StephenWeatherford:
How many users are using comments
How many users are using multi-line strings
What resourcesTypes/api versions are most used
$schema value used
apiVersion

### DIFF
--- a/src/DeploymentTemplate.ts
+++ b/src/DeploymentTemplate.ts
@@ -362,6 +362,17 @@ export class DeploymentTemplate {
         }
     }
 
+    public getMultilineStringCount(): number {
+        let count = 0;
+        this.visitAllReachableStringValues(jsonStringValue => {
+            if (jsonStringValue.unquotedValue.indexOf("\n") >= 0) {
+                ++count;
+            }
+        });
+
+        return count;
+    }
+
     public get jsonParseResult(): Json.ParseResult {
         return this._jsonParseResult;
     }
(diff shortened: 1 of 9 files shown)


src/DeploymentTemplate.ts
```diff
@@ -323,6 +334,70 @@ export class DeploymentTemplate {
...
+
+        return resourceCounts;
+
+        function traverseResources(resourcesObject: Json.ArrayValue, parentKey: string | undefined): void {
+            for (let resource of resourcesObject.elements) {
+                const resourceObject = Json.asObjectValue(resource);
+                if (resourceObject) {
+                    const resourceType = Json.asStringValue(resourceObject.getPropertyValue(templateKeys.resourceType));
+                    const apiVersion = Json.asStringValue(resourceObject.getPropertyValue(templateKeys.resourceApiVersion));
+                    if (resourceType && apiVersion) {
```
jimmylewis: Can we also get data when apiVersion is blank?  I.e. when someone is using an apiProfile?

src/DeploymentTemplate.ts
```diff
@@ -323,6 +334,70 @@ export class DeploymentTemplate {
...
+                        const isTypeAnExpression = resourceType.unquotedValue.trim().startsWith('[');
```
jimmylewis: technically there are more rules to being an expression.  Is this a shortcut just for type/apiVersion since those would never otherwise have `[`?  Or should we be more precise about checking for an expression?

StephenWeatherford: Yeah, this is just a shortcut, it's unlikely the other rules would be used here.

src/DeploymentTemplate.ts
StephenWeatherford: Sure.

PR #369, opened 2019-11-01
Add telemetry for DT errors/warnings

StephenWeatherford:


src/AzureRMTools.ts
```diff
@@ -162,7 +162,8 @@ export class AzureRMTools {
                             return;
                         }
 
-                        // Telemetry for template opened
+                        // Template for template opened
```
bhsubra: Telemetry for template opened?

StephenWeatherford: I fixed that, must have slipped in through a merge.  Thx.

PR #371, opened 2019-11-01
Add setting to disable checking for current schema

StephenWeatherford:
Thinking this might be necessary, haven't run it by PM yet, but want to get the code review done in case the bug bash decides we want this.


StephenWeatherford:
Thanks, Bhavya.  AlexF is in agreement.

PR #387, opened 2019-11-09
Attempt to fix #315 PositionContext assertions

### LINKED ISSUE #315: Telemetry error instance greater in 0.7.0: documentColumnIndex (1) cannot be greater than the line's maximum index (0)

StephenWeatherford:
Hopefully fixes #315 

The most likely culprit is that the document is out of sync with the positions being given to us.

I've never actually able to repro it.

### DIFF
--- a/src/AzureRMTools.ts
+++ b/src/AzureRMTools.ts
@@ -108,7 +108,14 @@ export class AzureRMTools {
 
     private getDeploymentTemplate(document: vscode.TextDocument): DeploymentTemplate | undefined {
         assert(document);
-        return this._deploymentTemplates.get(document.uri.toString());
+        const dt = this._deploymentTemplates.get(document.uri.toString());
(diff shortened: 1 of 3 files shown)


src/AzureRMTools.ts
```diff
@@ -108,7 +108,14 @@ export class AzureRMTools {
 
     private getDeploymentTemplate(document: vscode.TextDocument): DeploymentTemplate | undefined {
         assert(document);
-        return this._deploymentTemplates.get(document.uri.toString());
+        const dt = this._deploymentTemplates.get(document.uri.toString());
+        if (dt && dt.documentVersion !== document.version) {
+            // tslint:disable-next-line: no-console
+            console.log("Document version does not match cached version");
+            return undefined;
```
StephenWeatherford: Returning undefined - is right thing to do when calling updateDeploymentTemplate?

PR #415, opened 2019-12-12
Adding reload schemas command

alexgav:


src/languageclient/ReloadSchemasRequest.ts
```diff
@@ -0,0 +1,4 @@
+export class ReloadShemasRequest {
```
StephenWeatherford: an interface would make more sense here, don't actually need a runtime concept for it:

export interface IReloadSchemasRequest { 
}

src/languageclient/reloadSchemas.ts
```diff
@@ -0,0 +1,19 @@
...
+import { ext } from '../extensionVariables';
+import { ReloadShemasRequest } from './ReloadSchemasRequest';
+
+export async function reloadSchemas(): Promise<void> {
+
+    ext.outputChannel.appendLine("Reloading schemas...");
+
+    if (ext.languageServerClient) {
+        const request = new ReloadShemasRequest();
+        await ext.languageServerClient.sendRequest<ReloadShemasRequest>('schema/reloadSchemas', request);
```
StephenWeatherford: With IReloadSchemasRequest as an interface, I'd just do:

await ext.languageServerClient.sendRequest<IReloadShemasRequest>('schema/reloadSchemas', {});

src/languageclient/startArmLanguageServer.ts
```diff
@@ -27,15 +27,13 @@ export enum LanguageServerState {
...
 export async function stopArmLanguageServer(): Promise<void> {
     ext.outputChannel.appendLine(`Stopping ${languageServerName}...`);
     // Work-around for https://github.com/microsoft/vscode/issues/83254 - store languageServerState global via ext to keep it a singleton
     ext.languageServerState = LanguageServerState.Stopped;
-    if (client) {
-        await client.stop();
-        client = undefined;
+    if (ext.languageServerClient) {
+        await ext.languageServerClient.stop();
+        ext.languageServerClient = undefined;
```
StephenWeatherford: These two lines should probably be reversed.

src/languageclient/reloadSchemas.ts
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    ext.outputChannel.appendLine("Schemas reloaded");
```
StephenWeatherford: The schemas won't actually be reloaded at this point (you simply sent a notification to the lang server, right? so await will just wait until the notification is sent, maybe not even received).

How about getting rid of the "Reloading schemas..." and putting "Request to reload schemas sent" or something?

src/languageclient/startArmLanguageServer.ts
StephenWeatherford: (using a temp var) - and I realize, not your code :-)

src/languageclient/reloadSchemas.ts
```diff
@@ -0,0 +1,19 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { ext } from '../extensionVariables';
+import { ReloadShemasRequest } from './ReloadSchemasRequest';
+
+export async function reloadSchemas(): Promise<void> {
+
+    ext.outputChannel.appendLine("Reloading schemas...");
+
+    if (ext.languageServerClient) {
```
StephenWeatherford: How should we handle when the language server hasn't started yet?  It should be the same as when the lang server has started but isn't ready yet (you'll get an error sending the notification before the await ext.languageServerClient.onReady() line has run).  I'd say two possibilities:
1) throw an error, tell the user to wait until lang server is ready
2) set a flag to send the notification as soon as it's ready

I'm okay with either.  Ideally, I think we should wait to assign ext.languageServerClient until after the await onReady call completes, though, so we avoid this ambiguity - we either have a lang server and it's ready, or we don't have it yet.

StephenWeatherford: I'm also hesitating on whether to expose ext.languageServerClient.  I'm thinking maybe instead exposing a sendLanguageServerNotification<T>(type, args, waitForStart: boolean) function from the startArmLanguageServer.ts module.  Thoughts?

src/languageclient/reloadSchemas.ts
alexgav: I'm not sure how this works in TypeScript/JS - in C# the code would not resume until the request is completed. It would continue if you didn't await and it did a task discard. If it works differently in JavaScript/TS, I'll take your word for it.

alexgav: Could you take a look at this? 
https://stackoverflow.com/questions/43302584/why-doesnt-the-code-after-await-run-right-away-isnt-it-supposed-to-be-non-blo

alexgav: Oh, I see, I'm not sure if await here will actually wait for the response or not. Let me see if I can find documentation on that.

StephenWeatherford: Not about TS, it's about your implementation of ISchemaReloadHandler in the lang server as IJsonRpcNotificationHandler<SchemaReloadParams>.  I mentioned in the code review I wasn't sure that would wait for the handler to finish.  Looking at it again, I think it might.  You should just try and see if it waits or not...

src/languageclient/reloadSchemas.ts
alexgav: I was following Razor model (more or less). The do wrap the LanguageClient field into RazorLanguageServerClient, but they do expose an instance of RazorLanguageServerClient to the rest of the extension. I'd be OK either way.

src/languageclient/startArmLanguageServer.ts
alexgav: Sure, that makes sense, since we now have external users.

src/languageclient/reloadSchemas.ts
alexgav: Would we even need the interface then? I wasn't sure if we needed that parameter type on the server side. I simply didn't see any examples not using any, and the documentation isn't great - so I was just following the copy/modify routine :)

src/languageclient/ReloadSchemasRequest.ts
alexgav: Do we need anything at all then? Doesn't seem like that interface would get used. Let me try without it and see if things still work.

alexgav: I'm not sure how RPC / serialization / deserialization works here. I was trying to match parameter class name in case it's significant.

package.json
```diff
@@ -43,7 +43,8 @@
     "onLanguage:arm-template",
     "onLanguage:json",
```
StephenWeatherford: Should we put an editor context menu item to reload the schemas like in VS?  Just add this to the package.json:

        "menus": {
            "editor/context": [{
                "command": "azurerm-vscode-tools.reloadSchemas",
                "when": "editorLangId==arm-template"
            }]
        },

src/languageclient/ReloadSchemasRequest.ts
StephenWeatherford: I'd be very surprised if it did.  TS is a duck-typed language, so we tend to use classes much less.

src/languageclient/reloadSchemas.ts
StephenWeatherford: I don't believe I used an interface at all in my sample I sent you.  It's not a bad idea to have it - it will add intellisense and compile-time checking to the arg parameter, but my guess is it will work without it.

src/languageclient/reloadSchemas.ts
StephenWeatherford: I'm not too worried about it either way for this change, although it if were me I'd probably not expose the lang server client itself since it's really an implementation detail.  Either I'd expose an interface with specific fields or more likely just expose functions from the module.

src/languageclient/ReloadSchemasRequest.ts
alexgav: OK, the name doesn't seem to matter, and we don't need that interface or the type. I'll delete it in the next commit.

src/languageclient/reloadSchemas.ts
alexgav: We don't seem to need the interface. I'm deleting it.

src/languageclient/reloadSchemas.ts
alexgav: I double-checked, and as I thought, awaiting on the promise from the sendRequest won't execute any code that come after that. You can try setting a breakpoint on the server or adding await Task.Delay in the server code to verify the client code won't continue executing after await until the server request actually returns. I'm not attached to having two messages, let me know if you still would prefer to have just one - totally fine with that.

StephenWeatherford: Then in that case it's only a question of...  A bit pickier...  At that point will the schemas actually have been downloaded and reloaded, or just the cache cleared and the downloads will happen afterwards?  If the former, we're good.  If the latter, might consider a change in second message.  Thx.

package.json
alexgav: Sounds good to me.

src/languageclient/reloadSchemas.ts
alexgav: I left it as is for now since we are a bit pressed for time, but we can chat later and clean up as we see fit, possibly creating our own ArmLanguageServerClient class encapsulating LanguageClient kind of like Razor does.

StephenWeatherford: Then I'd suggest you throw an error if langserver==null rather than just doing nothing.

src/languageclient/reloadSchemas.ts
alexgav: Good point, schemas are just deleted at that point and the download isn't actually completed. I think I'll just go with a single message for simplicity sake.

src/languageclient/reloadSchemas.ts
```diff
@@ -4,16 +4,16 @@
...
-
     if (ext.languageServerClient) {
-        const request = new ReloadShemasRequest();
-        await ext.languageServerClient.sendRequest<ReloadShemasRequest>('schema/reloadSchemas', request);
+        ext.outputChannel.appendLine("Reloading schemas");
+
+        await ext.languageServerClient.sendRequest('schema/reloadSchemas', {});
+    }
+    else {
+        ext.outputChannel.appendLine("Language server is not yet ready");
```
StephenWeatherford: You really should throw an Error because they won't see this message unless they have the window open.  The error will be caught and displayed automatically.  We can improve later.

StephenWeatherford:
You need to fix the lint error.  Either:
1) command line or terminal: npm run lint
2) vscode: F1 -> Run Task -> npm lint

src/languageclient/reloadSchemas.ts
alexgav: Yep, had that change made already but didn't have it pushed. Pushed now.

PR #419, opened 2019-12-13
Add diagnostic links to errors

StephenWeatherford:


src/util/wrapError.ts
```diff
@@ -0,0 +1,38 @@
...
+ * ------------------------------------------------------------------------------------------ */
+
+import * as os from 'os';
+import { parseError } from 'vscode-azureextensionui';
+
+/**
+ * "Wraps" an existing throwable item, placing a new message at the beginning of the current message.
+ * Retains original stack info and other properties
+ * @param innerError Original error or other item that was thrown
+ * @param wrappingMessage Message to wrap the original message with
```
alexgav: Nit: wrong parameter name for wrappingMessage/outerMessage.

src/util/wrapError.ts
```diff
@@ -0,0 +1,38 @@
...
+    } else {
+        return new Error(newMessage);
+    }
+}
+
+function cloneError(error: Error): Error {
+    const copy = new Error();
+    for (const propName of Object.getOwnPropertyNames(error)) {
+        try {
+            const value = error[propName];
```
alexgav: Why do we need a temp variable?

PR #442, opened 2020-01-10
Enable strict mode

StephenWeatherford:
Engineering work, fixes the remaining issues that were keeping us from full strict mode.  This was done in three phases (see commits).

### DIFF
--- a/src/TLE.ts
+++ b/src/TLE.ts
@@ -46,7 +46,7 @@ export function asPropertyAccessValue(value: Value | null): PropertyAccess | nul
  * The Value class is the generic base class that all other TLE values inherit from.
  */
 export abstract class Value {
-    private _parent: ParentValue | null; // should be in constructor?
+    private _parent: ParentValue | null = null; // CONSIDER: should be in constructor?
 
     public get parent(): ParentValue | null {
         return this._parent;
@@ -1090,14 +1090,11 @@ export class ParseResult {
  * A TLE tokenizer that generates tokens from a TLE string.
  */
 export class Tokenizer {
-    private _basicTokenizer: basic.Tokenizer;
-    private _text: string;
-
-    private _current: Token | null;
+    private _current: Token | null = null;
     // This offset (+1) is because we trimmed off the initial quote character.
     private _currentTokenStartIndex: number = 1;
 
-    private constructor() {
+    private constructor(private _basicTokenizer: basic.Tokenizer, private _text: string) {
     }
 
     public static fromString(stringValue: string): Tokenizer {
@@ -1109,9 +1106,7 @@ export class Tokenizer {
         const trimmedLength: number = stringValue.length - (stringValue.endsWith(initialQuoteCharacter) ? 2 : 1);
         const trimmedString: string = stringValue.substr(1, trimmedLength);
 
-        const tt = new Tokenizer();
-        tt._basicTokenizer = new basic.Tokenizer(trimmedString);
-        tt._text = stringValue;
+        const tt = new Tokenizer(new basic.Tokenizer(trimmedString), stringValue);
         return tt;
     }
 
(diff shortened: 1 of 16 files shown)


StephenWeatherford:
ping

src/TLE.ts
```diff
@@ -1252,15 +1247,13 @@ export class Token {
     private _span: language.Span;
     private _stringValue: string;
 
-    private static create(tokenType: TokenType, startIndex: number, stringValue: string): Token {
+    public constructor(tokenType: TokenType, startIndex: number, stringValue: string) {
```
alexgav: >tokenType: TokenType, startIndex: number, stringValue: string [](start = 23, length = 61)

Nit: Why not use property definition via constructor args here live you've done in the Tokenizer constructor above? I don't have strong feelings about it, but it seems it would be good to be consistent in style.

src/TLE.ts
```diff
@@ -1280,61 +1273,61 @@ export class Token {
...
-        return Token.create(TokenType.Comma, startIndex, ",");
+        return new Token(TokenType.Comma, startIndex, ",");
     }
 
     public static createPeriod(startIndex: number): Token {
-        return Token.create(TokenType.Period, startIndex, ".");
+        return new Token(TokenType.Period, startIndex, ".");
     }
 
     public static createWhitespace(startIndex: number, stringValue: string): Token {
```
alexgav: >createWhitespace [](start = 18, length = 16)

Nit: unrelated to this PR, seems like this should've asserted that stringValue is actually whitespace, since other createXXX functions check the value

src/TLE.ts
```diff
@@ -1280,61 +1273,61 @@ export class Token {
...
         assert(1 <= stringValue.length);
 
-        return Token.create(TokenType.Whitespace, startIndex, stringValue);
+        return new Token(TokenType.Whitespace, startIndex, stringValue);
     }
 
     public static createQuotedString(startIndex: number, stringValue: string): Token {
         assert(stringValue != null);
         assert(1 <= stringValue.length);
         assert(Utilities.isQuoteCharacter(stringValue[0]));
```
alexgav: >assert(Utilities.isQuoteCharacter(stringValue[0])); [](start = 8, length = 51)

Nit: Unrelated to this PR, why not check the last character as well?

src/TLE.ts
```diff
@@ -1280,61 +1273,61 @@ export class Token {
...
 
-        return Token.create(TokenType.QuotedString, startIndex, stringValue);
+        return new Token(TokenType.QuotedString, startIndex, stringValue);
     }
 
     public static createNumber(startIndex: number, stringValue: string): Token {
         assert(stringValue != null);
         assert(1 <= stringValue.length);
         assert(stringValue[0] === "-" || Utilities.isDigit(stringValue[0]));
```
alexgav: >Utilities.isDigit(stringValue[0]) [](start = 41, length = 33)

Nit: Unrelated to this PR, why check only the first character? What if this was "1z"?

src/extensionVariables.ts
```diff
@@ -7,29 +7,84 @@ import * as vscode from "vscode";
...
 import { isWebpack } from "./constants";
+import { assert } from "./fixed_assert";
 import { LanguageServerState } from "./languageclient/startArmLanguageServer";
 import { JsonOutlineProvider } from "./Treeview";
 
+/**
+ * Represents a scalar value that must be initialized before its getValue is called
+ */
+class InitializeBeforeUse<T> {
+    private _value: { value: T; initialized: true } | { initialized: false } = { initialized: false };
```
alexgav: >private _value: { value: T; initialized: true } | { initialized: false } = { initialized: false }; [](start = 4, length = 98)

Nit: I'd probably use similar technique for CachedValue, or make this class more similar to CachedValue for consistency :) It's interesting the stuff TS lets you do.

src/TLE.ts
StephenWeatherford: The token will be created even if the string is malformed, e.g. no ending quote.  Not sure that's the optimal design, but it's the way it's always been.

src/TLE.ts
StephenWeatherford: I'd have to check to be sure, but pretty sure the validation is done before calling this function.  Agreed not great design, but probably not worth the change by itself right now.

src/extensionVariables.ts
StephenWeatherford: I discovered this trick more recently. :-)

Actually not happy with how I'm having to do this class, but it's partly due to a bug in vscode.

src/TLE.ts
StephenWeatherford: Not too worried on consistency for that.  I tend to use it for new code, but none of the old code has it.  Changing over everything would be a lot of work.

StephenWeatherford:
Thanks!

PR #443, opened 2020-01-10
Replace usage of null with undefined

StephenWeatherford:
Currently a lot of the old code uses null, while most of the new code uses undefined.  Experience has shown there's not really an advantage to using both, and our other extensions all use undefined.  Now that strict null checking has been fully enabled, I've switched the null code to use undefined.

Lots of changes in multiple phases, supported by the tests passing.  Don't really expect any problems, don't think this warrants much review.

Note: builds on https://github.com/microsoft/vscode-azurearmtools/pull/442


StephenWeatherford:
ping

src/AzureRMAssets.ts
```diff
@@ -224,7 +227,7 @@ interface FunctionMetadataContract {
         expectedUsage: string;
         description: string;
         minimumArguments: number;
-        maximumArguments: number;
+        maximumArguments: number | null;
```
alexgav: >number | null; [](start = 26, length = 14)

I'm confused - I thought we were switching from null to undefined? Or does it have to be null here?

alexgav:
>         assert(_value !== undefined);

We do want to allow the value to be undefined?

---
Refers to: src/JSON.ts:912 in c0ea7af. [](commit_id = c0ea7afe13c1450c0b7014849c0e69bfe861e287, deletion_comment = True)

alexgav:
>                     if (tleValue.leftSquareBracketToken !== null && tleValue.leftSquareBracketToken.span.startIndex === tleCharacterIndex) {

How come we don't need this check anymore? We check it above in if, but here we are in the else case of that if. Shouldn't we need to check it here as well?

---
Refers to: src/TLE.ts:580 in c0ea7af. [](commit_id = c0ea7afe13c1450c0b7014849c0e69bfe861e287, deletion_comment = True)

src/visitors/IncorrectFunctionArgumentCountVisitor.ts
```diff
@@ -59,7 +60,7 @@ export class IncorrectFunctionArgumentCountVisitor extends Visitor {
 
             minimumArguments = functionMetadata.minimumArguments;
             // tslint:disable-next-line:max-line-length
-            assert(minimumArguments !== null && minimumArguments !== undefined, `TLE function metadata for '${actualFullFunctionName}' has a null or undefined minimum argument value.`);
+            assert(!isNullOrUndefined(minimumArguments), `TLE function metadata for '${actualFullFunctionName}' has a null or undefined minimum argument value.`);
```
alexgav: > assert(!isNullOrUndefined(minimumArguments), `TLE function metadata for '${actualFullFunctionName}' has a null or undefined minimum argument value.`); [](start = 11, length = 151)

So I'm kind of confused here. We removed "| undefined" from minimumArguments declaration. If functionMetadata.minimumArguments still has "| undefined" declared, why is this not causing a compile error? And do we need to check minimumArguments for both null and undefined?

src/PositionContext.ts
```diff
@@ -64,38 +65,35 @@ export class PositionContext {
...
+    private constructor(deploymentTemplate: DeploymentTemplate) {
+        this._deploymentTemplate = deploymentTemplate;
+    }
 
     public static fromDocumentLineAndColumnIndexes(deploymentTemplate: DeploymentTemplate, documentLineIndex: number, documentColumnIndex: number): PositionContext {
-        assert(deploymentTemplate !== null, "deploymentTemplate cannot be null");
-        assert(deploymentTemplate !== undefined, "deploymentTemplate cannot be undefined");
-        assert(documentLineIndex !== null, "documentLineIndex cannot be null");
-        assert(documentLineIndex !== undefined, "documentLineIndex cannot be undefined");
+        assert(deploymentTemplate);
```
alexgav: Why do we not need !isNullOrUndefined here?

src/PositionContext.ts
```diff
@@ -64,38 +65,35 @@ export class PositionContext {
...
         context._givenDocumentPosition = new language.Position(documentLineIndex, documentColumnIndex);
         return context;
 
     }
     public static fromDocumentCharacterIndex(deploymentTemplate: DeploymentTemplate, documentCharacterIndex: number): PositionContext {
-        assert(deploymentTemplate !== null, "deploymentTemplate cannot be null");
-        assert(deploymentTemplate !== undefined, "deploymentTemplate cannot be undefined");
-        assert(documentCharacterIndex !== null, "documentCharacterIndex cannot be null");
-        assert(documentCharacterIndex !== undefined, "documentCharacterIndex cannot be undefined");
+        assert(!isNullOrUndefined(deploymentTemplate), "deploymentTemplate cannot be null/undefined");
```
alexgav: >isNullOrUndefined [](start = 16, length = 17)

Why are we using !isNullOrUndefined on deploymentTemplate here but not in fromDocumentLineAndColumnIndexes above?

src/Language.ts
```diff
@@ -146,11 +145,9 @@ export enum IssueKind {
  */
 export class Issue {
     constructor(private _span: Span, private _message: string, public kind: IssueKind) {
-        assert(_span !== null, "_span must not be null.");
-        assert(_span !== undefined, "_span must not be undefined.");
+        assert(_span, "_span must not be null/undefined");
```
alexgav: >_span [](start = 15, length = 5)

Why do we not need !IsNullOrUndefined here?

src/ReferenceList.ts
```diff
@@ -44,8 +44,7 @@ export class ReferenceList {
     }
 
     public translate(movement: number): ReferenceList {
-        assert(movement !== null, "Cannot translate a reference list by a null amount.");
-        assert(movement !== undefined, "Cannot translate a reference list by an undefined amount.");
+        assert(movement);
```
alexgav: >movement [](start = 15, length = 8)

Why do we not need !isNullOrUndefined here?

test/support/assertNotNull.ts
```diff
@@ -5,9 +5,12 @@
 
 'use strict';
 
-export function assertNotNull<T>(v: T | undefined | null): T {
+export function assertNotNull<T>(v: T | undefined | null, argNameOrMessage?: string): T {
```
alexgav: >assertNotNull [](start = 16, length = 13)

Should this be called assertNotNullOrUndefined?

src/DeploymentTemplate.ts
```diff
@@ -58,9 +59,8 @@ export class DeploymentTemplate {
      * @param _documentId A unique identifier for this document. Usually this will be a URI to the document.
      */
     constructor(private _documentText: string, private _documentId: string) {
-        assert(_documentText !== null);
-        assert(_documentText !== undefined);
-        assert(_documentId);
```
alexgav: >_documentId [](start = 17, length = 11)

How come we don't need to check _documentId anymore?

src/JSON.ts
```diff
@@ -904,12 +905,9 @@ export class NullValue extends Value {
...
     private readonly _debugText: string; // Used only for debugging - copy of the original text being parsed
 
-    constructor(private _tokens: Token[], private _lineLengths: number[], private _value: Value | null, text: string, public readonly commentCount: number) {
-        assert(_tokens !== null);
-        assert(_tokens !== undefined);
-        assert(_lineLengths !== null);
-        assert(_lineLengths !== undefined);
-        assert(_value !== undefined);
+    constructor(private _tokens: Token[], private _lineLengths: number[], private _value: Value | undefined, text: string, public readonly commentCount: number) {
+        nonNullValue(_tokens, "_tokens");
```
alexgav: >_tokens [](start = 31, length = 7)

Nit: For the future, it would be nice to have some sort of nameof functionality, like [https://stackoverflow.com/questions/50470025/nameof-keyword-in-typescript](https://stackoverflow.com/questions/50470025/nameof-keyword-in-typescript)

src/Language.ts
alexgav: OK, you changed it later, nevermind.

---
In reply to: [366153737](https://github.com/microsoft/vscode-azurearmtools/pull/443#discussion_r366153737) [](ancestors = 366153737)

src/util/nonNull.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export function nonNullProp<TSource, TKey extends keyof TSource>(source: TSource, name: TKey): NonNullable<TSource[TKey]> {
+    const value: NonNullable<TSource[TKey]> = <NonNullable<TSource[TKey]>>source[name];
+    return nonNullValue(value, <string>name);
+}
+
+/**
+ * Validates that a given value is not null and not undefined.
+ */
+export function nonNullValue<T>(value: T | undefined, propertyNameOrMessage?: string): T {
```
alexgav: >nonNullValue [](start = 16, length = 12)

Nit: Since null and undefined are different, would it make more sense to make it verbose and specify nonNullOrUndedinedValue to be more explicit here? Same for other methods in this file.

alexgav: Nit: Also, looking at usages, we are essentially using side-effect of the getter to assert/throw... I personally would've probably broke the check + throw into a separate "ThrowIfNullOrUndefined" helper that would get used in both these getters and outside of the getters when you just want to check the value. Seems a bit cleaner to me, but not a big deal.

---
In reply to: [366155874](https://github.com/microsoft/vscode-azurearmtools/pull/443#discussion_r366155874) [](ancestors = 366155874)

src/AzureRMAssets.ts
StephenWeatherford: It has to be null here because JSON doesn't support undefined, and this is the contract for reading the metadata from JSON.

StephenWeatherford:
> > ```
> >     assert(_value !== undefined);
> > ```
> 
> 
> We do want to allow the value to be undefined?

Yes - the JSON could be blank or only comments, therefore no top-level JSON value. Before it was null for that case, now it will be undefined instead.

StephenWeatherford:
> > ```
> >                 if (tleValue.leftSquareBracketToken !== null && tleValue.leftSquareBracketToken.span.startIndex === tleCharacterIndex) {
> > ```
> 
> 
> How come we don't need this check anymore? We check it above in if, but here we are in the else case of that if. Shouldn't we need to check it here as well with !!tleParseResult.leftSquareBracketToken ?
> 
> 
> 
> Refers to: src/TLE.ts:580 in [c0ea7af](https://github.com/microsoft/vscode-azurearmtools/commit/c0ea7afe13c1450c0b7014849c0e69bfe861e287). [](commit_id = [c0ea7af](https://github.com/microsoft/vscode-azurearmtools/commit/c0ea7afe13c1450c0b7014849c0e69bfe861e287), deletion_comment = True)

The first was TLE.ParseResult.leftSquareBracket (defined as Token|undefined), the second was ArrayAccessValue.leftSquareBracketToken (defined as Token).  Same name, different classes.

src/visitors/IncorrectFunctionArgumentCountVisitor.ts
StephenWeatherford: Good question.  I've declared the contract and the json schema for the metadata such that minimumArguments is number (not null). So theoretically it shouldn't be either.  But this does come from reading a JSON file, so I can't entirely guarantee it, esp. if we choose to read this file from the cloud later (we used to do that), so I'd rather still have the check in there.  I don't really like comparing against null and undefined directly, because usually we just check against both of them at the same time, using a falsy check - assert(minimumArguments). But here that's problematic because it could be the valid falsy value of zero. I just prefer the isNullOrUndefined way rather than two separate checks.

src/PositionContext.ts
StephenWeatherford: Fixed

src/ReferenceList.ts
StephenWeatherford: Good catch, can't use a simple falsy check because it could theoretically be zero.  That's why I added isNullOrUndefined, to make these cases easier. (Have I mentioned that JavaScript is a horribly-designed language?).  Now using nonNullValue.

src/DeploymentTemplate.ts
StephenWeatherford: Another good catch, surprised the suites didn't catch that.  Meant this:
```
nonNullOrEmptyValue(_documentId, "_documentId");
```
We don't actually need most of the null/undefined checks - they're left over from when the compiler couldn't check for null/undefined. I've just decided to leave them in for now, but I guess I'm not really consistent with it.

src/util/nonNull.ts
StephenWeatherford: > > nonNullValue [](start = 16, length = 12)
> 
> Nit: Since null and undefined are different, would it make more sense to make it verbose and specify nonNullOrUndedinedValue to be more explicit here? Same for other methods in this file.

Yes, but I pulled it from another of our repros and was loathe to change the name. Also, since we don't normally handle null/undefined differently, the explicit/wordy naming seemed less important (this function is in other repos frequently used inside expressions, and the longer name makes the code harder to read).

StephenWeatherford: > Nit: Also, looking at usages, we are essentially using side-effect of the getter to assert/throw... I personally would've probably broke the check + throw into a separate "ThrowIfNullOrUndefined" helper that would get used in both these getters and outside of the getters when you just want to check the value. Seems a bit cleaner to me, but not a big deal.
> 
> In reply to: [366155874](https://github.com/microsoft/vscode-azurearmtools/pull/443#discussion_r366155874) [](ancestors = 366155874)

Related to the other question.  The main usage scenario in our other repos is this:
```
const myvalue = someexpression(nonNullProp(object, "propName"))
```

This was needed mostly because of some poorly-typed SDK objects.  I considered adding  assertIfNotX() but since they're completely equivalent decided not to bother.  This seems just as readable. But likewise no strong opinion.

src/JSON.ts
StephenWeatherford: Yes, have been wishing for that.  The nonNullProp is the closest I was able to come - kind of solves the problem for object properties, but doesn't help with method parameters.  I'll have to look at https://github.com/dsherret/ts-nameof, but sounds a little expensive.

StephenWeatherford:
Thanks!

PR #448, opened 2020-01-15
Add survey

StephenWeatherford:


src/survey.ts
```diff
@@ -0,0 +1,212 @@
...
+
+import { commands, MessageItem, Uri, window, workspace } from 'vscode';
+import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azureextensionui";
+import { configPrefix } from './constants';
+import { ext } from "./extensionVariables";
+import { assert } from './fixed_assert';
+import { httpGet } from "./httpGet";
+import { hoursToMs, minutesToMs, weeksToMs } from "./util/time";
+
+namespace stateKeys {
```
alexgav: >stateKeys [](start = 10, length = 9)

Nit: should we call this surveryStateKeys to disambiguate from possible other stateKeys we may have?

src/survey.ts
```diff
@@ -0,0 +1,212 @@
...
+     */
+    export function registerActiveUse(): void {
+        // Don't wait
+        // tslint:disable-next-line: no-floating-promises
+        callWithTelemetryAndErrorHandling("considerShowingSurvey", async (context: IActionContext) => {
+            context.errorHandling.suppressDisplay = true;
+            // This gets called a lot, we don't want telemetry for most of it
+            context.telemetry.suppressIfSuccessful = true;
+
+            if (isReentrant) {
```
alexgav: >isReentrant [](start = 16, length = 11)

Do we need to lock before checking / setting this? If the user opens multiple documents, can this be called multiple times on different threads?

src/survey.ts
StephenWeatherford: It's limited to this file so not an issue.

src/survey.ts
StephenWeatherford: JavaScript is single-threaded.  Much easier. :-)

src/survey.ts
```diff
@@ -0,0 +1,212 @@
...
+    export const surveyPostponedUntilTime = 'surveyPostponedUntilTime';
+}
+
+const linkToSurvey = 'https://aka.ms/arm-tools-survey';
+const surveyPrompt = "Could you please take 2 minutes to tell us how well the Azure Resource Manager Tools extension is working for you?";
+
+// Wait for this amount of time since extension first used (this session) before
+// considering asking the survey (only if still interacting with extension)
+interface ISurveyConstants {
+    activeUsageTimeBeforeSurvey: number;
```
alexgav: >activeUsageTimeBeforeSurvey [](start = 4, length = 27)

Nit: maybe add Ms suffix to these, though it's kind of obvious from the usage immediately below.

Issue #449, opened 2020-01-21
Snippets do not include scaffolding for Subscription, Management Group, or Tenant deployments


neilpeterson:
Current snippets only include scaffolding for resource group scoped deployments.

![image](https://user-images.githubusercontent.com/7844635/72842237-d19b0500-3c4c-11ea-8cdd-8422a4d1a675.png)

```
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {},
    "variables": {},
    "resources": [],
    "outputs": {},
    "functions": []
}
```

PR #450, opened 2020-01-21
Schema scope

neilpeterson:
Updated deployment and parameter snippet schemas:

- Added snippet for Subscription, Management Group, and Tenant deployments
- Updated Resource Group deployment schema version to 2019-04-01
- Updated parameter schema version to 2019-04-01
- Renamed existing snippets (prefix) for clarity
- Updated existing descriptions for clarity

Sample:

image

Fixes https://github.com/microsoft/vscode-azurearmtools/issues/449


nilshedstrom:
I like that you can create a new arm template with "arm!" and a new parameter file with "armp!"
I would prefer to stick with the current ones compared to "arm-deploy-file-rg"

neilpeterson:
Thanks @nilshedstrom I also agree that the muscle memory / quick keystroke on `arm!` is nice, but it's not very descriptive, particularly after adding the additional scopes. Do you have any thoughts on how to handle when adding the new scopes (subscription, management group, and tenant)?

assets/jsonsnippets.jsonc
```diff
@@ -12,18 +12,63 @@
...
+            "    \"\\$schema\": \"https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#\",",
+            "    \"contentVersion\": \"1.0.0.0\",",
+            "    \"parameters\": {},",
+            "    \"variables\": {},",
+            "    \"resources\": [],",
+            "    \"outputs\": {},",
+            "    \"functions\": []",
+            "}"
+        ],
+        "description": "Tanant Deployment"
```
StephenWeatherford: tenant

StephenWeatherford:
Here's what it looks like with your current changes:
image
Here are a couple other possibilities:
image
image

nilshedstrom:
> image
@neilpeterson I like the last suggestion from @StephenWeatherford

assets/jsonsnippets.jsonc
neilpeterson: Thanks, addressed.

neilpeterson:
Thanks for the suggestions @StephenWeatherford  and input @nilshedstrom. Let me gather some additional opinions and I will update.

neilpeterson:
@nilshedstrom / @StephenWeatherford option three wins :).

PR #469, opened 2020-02-03
Log telemetry from language server

alexgav:


StephenWeatherford:
When you add a PR to update the lang server nuget version, could you remove the alpha from the package.json version and re-run "npm i" to cause that change to go into package-lock.json as well?  Thx.

PR #480, opened 2020-02-08
Move to dotnet 3.0

StephenWeatherford:


src/acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -33,6 +33,7 @@ export class DotnetCoreAcquisitionWorker {
         '2.0': '2.0.9',
         2.1: '2.1.11',
         2.2: '2.2.5',
```
alexgav: Nit: should we make all property name strings for consistency?

PR #495, opened 2020-02-21
Use a single output channel

StephenWeatherford:
Currently there's a separate output channel for the extension and another for the language server.  That's confusing to customers, esp when we ask them to report what they see in the output window.


StephenWeatherford:
@alexgav ping

PR #498, opened 2020-02-22
Adding child resources support

alexgav:


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to the "vscode-azurearmtools" extension will be documented in this file.
 
+## Version 0.8.5-alpha
```
StephenWeatherford: I think this next release should be 0.9.0.

neilpeterson:
@alexgav I've hacked together a build to test out the child resource fix. To do so, I've updated the `ARM_LANGUAGE_SERVER_NUGET_VERSION` config in the extension to equal `3.0.0-preview.2012.3` and then created a new .vsix file. It looks to have worked because I do see an improvement in the child resource api version mis/match issue.

Note here that in the child resource I do see multiple resource versions where with the released extension we would only see the parent match.

image

However, when selecting the non-matched API version, the resource fails to validate with an indication that the api version must be the same as the parent.

image

If I do update the child resource api version to match, everything looks good.

Perhaps the steps I took to replicate the build are at fault here, but wanted to run this by you for review.

Thanks a bunch for working on this.

StephenWeatherford:
@alexgav @neilpeterson I merged this before noticing Neil's comments above.  I'm going to add a link to them in the existing bug.  Alex, can you investigate?  Thanks.

Also - how are we with testcases?  Ideally we should add some to the extension functional tests.

PR #510, opened 2020-03-03
ARM Snippet Updates (a - c)

neilpeterson:
Fixes scaffolding issues, snippet validation (a - c), brings resources to latest known schema version,  and fixes formatting issues.


assets/armsnippets.jsonc
```diff
@@ -838,15 +841,13 @@
...
+            "       \"[resourceId('Microsoft.DocumentDB/databaseAccounts/', '${1:account-name}')]\"",
             "    ],",
             "    \"properties\": {",
             "        \"resource\": {",
             "            \"id\": \"${2:database-name}\"",
             "        },",
-            "        \"options\": {",
-            "            \"throughput\": ${3:400}",
-            "        }",
+            "        \"options\": {}",
```
StephenWeatherford: Throughput can be quite expensive for cosmos, I think we should leave that here.  Or was there a specific reason you left it out?

assets/armsnippets.jsonc
```diff
@@ -856,17 +857,15 @@
...
-            "         }",
+            "     \"type\": \"Microsoft.DocumentDB/databaseAccounts/apis/keyspaces\",",
+            "     \"name\": \"[concat('${1:account-name}', '/cassandra/', '${2:database-name}')]\",",
+            "     \"apiVersion\": \"2016-03-31\",",
+            "     \"dependsOn\": [ \"[resourceId('Microsoft.DocumentDB/databaseAccounts/', '${1:account-name}')]\" ],",
+            "     \"properties\":{",
+            "         \"resource\":{",
+            "             \"id\": \"${3:keyspace}\"",
+            "         },",
+            "      \"options\":{}",
```
StephenWeatherford: ditto

assets/armsnippets.jsonc
```diff
@@ -967,7 +966,8 @@
             "                       \"path\": \"/${9:PathToNotIndex}/*\"",
             "                }]",
             "            }",
-            "        }",
+            "        },",
```
StephenWeatherford: Can you create a cosmos graph in the portal and compare?  They've made some important changes since GA'ing that likely aren't in here.

assets/armsnippets.jsonc
```diff
@@ -985,9 +985,7 @@
             "               \"resource\":{",
             "                   \"id\": \"${3:table-name}\"",
             "               },",
-            "            \"options\":{",
-            "                \"throughput\": ${4:400}",
-            "            }",
+            "            \"options\":{}",
```
StephenWeatherford: Not so this is so expensive but still probably an important option to keep front and center

assets/armsnippets.jsonc
```diff
@@ -1167,16 +1165,16 @@
...
-            "        \"kubernetesVersion\": \"${2:1.9}\",",
-            "        \"dnsPrefix\": \"$32:dnsPrefix}\",",
+            "        \"kubernetesVersion\": \"${2|1.15.7,1.15.5,1.14.8|}\",",
+            "        \"dnsPrefix\": \"${3:dnsprefix}\",",
             "        \"agentPoolProfiles\": [",
             "            {",
             "                \"name\": \"agentpool\",",
             "                \"count\": ${4:2},",
-            "                \"vmSize\": \"${5:virtualMachineSize}\",",
+            "                \"vmSize\": \"${5:Standard_A1}\",",
```
StephenWeatherford: Is it worth it to put in common values here?  I know they add a lot frequently, though, so maybe not??

neilpeterson: It's a pretty big list. I had considered putting in a few, however did not want to give the impression that the value was limited to that set.

assets/armsnippets.jsonc
neilpeterson: It was not validating as it was. I will need to dig into the resource, deploy a few, and wrap my head around how this should be configured.

StephenWeatherford: Did you try with quotes around it?  Sometimes the schemas are wrong about the types of things (or maybe its max is too large for an integer).  See https://github.com/Azure/azure-quickstart-templates/blob/master/101-cosmosdb-gremlin/azuredeploy.json.

neilpeterson: Yes, quotes fixes it. My conclusion is a schema issue. I am in the process of putting this back and then will follow up with an issue on the schema.

neilpeterson: Fixed

assets/armsnippets.jsonc
neilpeterson: Fixed

assets/armsnippets.jsonc
neilpeterson: Fixed

PR #511, opened 2020-03-04
ARM snippet updates (D - R)

neilpeterson:
Updated API versions to latest known
Updated snippets so that they pass validation
Renamed nested to linked
Updated mysql snippet so that it is functional
Reorder some for consistency (name, type, apiversion)


assets/armsnippets.jsonc
```diff
@@ -103,9 +103,9 @@
         "prefix": "arm-nested",
         "body": [
             "{",
-            "    \"name\": \"${1:nestedDeployment1}\",",
+            "    \"name\": \"${1:linkedDeployment1}\",",
```
StephenWeatherford: Have you looked up the official terminology?  Definitely seeing "nested templates" used for links, e.g. MS docs here: https://docs.microsoft.com/en-us/azure/lab-services/deploy-nested-template-environments.

assets/armsnippets.jsonc
```diff
@@ -1497,7 +1497,8 @@
             "        \"storageAccounts\": [",
             "            {",
             "                \"id\": \"[resourceId('Microsoft.Storage/storageAccounts', '${2:mediaServiceStorageAccount1}')]\",",
-            "                \"isPrimary\": true",
+            "                \"isPrimary\": true,",
```
StephenWeatherford: isPrimary/type=primary sound redundant??

assets/armsnippets.jsonc
```diff
@@ -1108,10 +1108,10 @@
             "        \"enabledForDeployment\": true,",
             "        \"enabledForTemplateDeployment\": true,",
             "        \"enabledForDiskEncryption\": true,",
-            "        \"tenantId\": \"${2:tenantId}\",",
+            "        \"tenantId\": \"${2:00000000-0000-0000-0000-000000000000}\",",
```
StephenWeatherford: Did you decide as a general rule we wouldn't mention variables?

assets/armsnippets.jsonc
neilpeterson: I was trying to set this up for an additional snippet. This one uses a tempalte link. We do not have a snippet that uses an inline tempalte. My thought was that one would be `arm-linked` and the other `arm-nested`. That said, I see here that I only renamed the variable and not resource name and not the snippet :).

I will revert this and then consider naming once we add the new snippet. Perhaps it will end up `arm-nested-linked` and `arm-nested-inline`.

assets/armsnippets.jsonc
neilpeterson: I replaced tenantId with a dummy value that validates, however I guess it's moot since this value will need to be updated. I will revert.

I would love to use variable / parameters in the snippets if we could also create the corresponding template variables / parameters.

assets/armsnippets.jsonc
neilpeterson: Fixed

assets/armsnippets.jsonc
neilpeterson: Fixed

assets/armsnippets.jsonc
neilpeterson: Fixed

PR #514, opened 2020-03-05
Support for params files

StephenWeatherford:
1.	Allow user to select a parameter file for each template file opened
	a.	Save in user settings, which also makes it available to the language server
	2.	Show parameter file in status bar (clicking it brings up Select Parameter File…)
	3.	When opening a template file, check for similarly-named parameter files. If found, ask user if they want to associate it.
	4.	Add "Reset global state" command (mostly for use with testing)

Depends on language server PR.


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "azurerm-vscode-tools",
     "displayName": "Azure Resource Manager (ARM) Tools",
     "description": "Language server, editing tools and snippets for Azure Resource Manager (ARM) template files.",
-    "version": "0.8.5-alpha",
+    "version": "0.9.0-alpha",
```
alexgav: >9 [](start = 18, length = 1)

Nit: Can we go past 1.0 without going GA?  Are we incrementing too much?

package.json
```diff
@@ -247,6 +290,30 @@
                     "when": "azurerm-vscode-tools.template-outline.active == true && viewItem == variables",
                     "group": "arm-template"
                 }
+            ],
+            "editor/title": [
+                {
+                    "command": "azurerm-vscode-tools.selectParameterFile",
+                    "group": "zzz_arm-template@2",
```
alexgav: >2 [](start = 47, length = 1)

Not @1 and @2?

src/AzureRMTools.ts
```diff
@@ -22,8 +23,10 @@ import * as Json from "./JSON";
 import * as language from "./Language";
 import { reloadSchemas } from "./languageclient/reloadSchemas";
 import { startArmLanguageServer, stopArmLanguageServer } from "./languageclient/startArmLanguageServer";
+import { considerQueryingForParameterFile, findMappedParamFileForTemplate, getFriendlyPathToParamFile, openParameterFile, selectParameterFile } from "./parameterFiles";
```
alexgav: >Parameter [](start = 128, length = 9)

Nit: A bit confusing that we use both "param" and "parameter". I'd stick with one or the other so it doesn't seem like there are some semantic differences (unless there are?)

src/AzureRMTools.ts
```diff
@@ -215,15 +227,13 @@ export class AzureRMTools {
...
-                                    this.queryUseNewerSchema(editor, deploymentTemplate);
-                                }
-                            }
+                            // Are they using an older schema?  Ask to update.
+                            // tslint:disable-next-line: no-suspicious-comment
+                            // TODO: Move to separate file
+                            this.considerQueryNewerSchema(editor, deploymentTemplate);
+
+                            // Is there a possibly-matching params file they might want to associate?
+                            considerQueryingForParameterFile(document);
```
alexgav: >Querying [](start = 36, length = 8)

Nit: naming consistency - I'd make both either Query or Querying

src/AzureRMTools.ts
```diff
@@ -492,6 +519,31 @@ export class AzureRMTools {
...
+        const activeDocument = vscode.window.activeTextEditor?.document;
+        if (activeDocument) {
+            const deploymentTemplate = this.getDeploymentTemplate(activeDocument);
+            if (deploymentTemplate) {
+                const paramFileUri = findMappedParamFileForTemplate(activeDocument.uri);
+                if (paramFileUri) {
+                    const doesParamFileExist = await fse.pathExists(paramFileUri?.fsPath);
+                    let text = `Parameters: ${getFriendlyPathToParamFile(activeDocument.uri, paramFileUri)}`;
+                    if (!doesParamFileExist) {
+                        text += " $(error) Not found";
```
alexgav: >(error) [](start = 35, length = 7)

Where does this come from?

src/acquisition/dotnetAcquisition.ts
```diff
@@ -76,7 +76,7 @@ export async function dotnetAcquire(
         }
         return await acquisitionWorker.acquire(version, telemetryProperties);
     } catch (err) {
-        const linkMessage = `>>>> This extension requires .NET Core for full functionality, but we were unable to download and install a local copy for the extension. If this error persists, please see https://aka.ms/vscode-armtools-dotnet for troubleshooting tips.`;
+        const linkMessage = `>>>> This extension requires .NET Core for full functionality, but we were unable to download and install a local copy for the extension. If this error persists, please see https://aka.ms/vscode-armtools-dotnet for troubleshooting tips. `;
```
alexgav: > ` [](start = 265, length = 2)

Intentional?

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+import { DeploymentTemplate } from './DeploymentTemplate';
+import { ext } from './extensionVariables';
+import { containsParamsSchema } from './schemas';
+
+const readAtMostBytesToFindParamsSchema = 4 * 1024;
+const currentMessage = "Current";
+const similarFilenameMessage = "Similar filename";
+const howToMessage = `You can manually associate a parameter file with this template at any time by clicking "Select Parameter File..." in the status bar or the editor context menu.`;
+
+const _filesToIgnoreThisSession: Set<string> = new Set<string>();
```
alexgav: >Set<string>(); [](start = 51, length = 14)

Nit: not sure if we should use case-insensitive file name set here on Windows (here and in other places we use sets of filenames)

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+  }
+  if (!sourceUri) {
+    await ext.ui.showWarningMessage(`No Azure Resource Manager template file is being edited.`);
+    return;
+  }
+
+  let templateUri: Uri = sourceUri;
+
+  // Verify it's a template file
+  const contents = (await fse.readFile(templateUri.fsPath, { encoding: "utf8" })).toString();
```
alexgav: >readFile [](start = 30, length = 8)

Would it make sense to read only to the limit we want to look at the file, i.e. 4K instead of reading the entire file?

package.json
StephenWeatherford: It goes to 0.10.0 next.

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+  const template: DeploymentTemplate = new DeploymentTemplate(contents, "Check file is template");
```
alexgav: >DeploymentTemplate [](start = 43, length = 18)

If we are using activeTextEditor, would there already be existing DeploymentTemplate so we don't have to check this again?

src/AzureRMTools.ts
StephenWeatherford: https://microsoft.github.io/vscode-codicons/dist/codicon.html

This is documented by vscode.

src/acquisition/dotnetAcquisition.ts
StephenWeatherford: Yeah, in the errors I was seeing no space after this before the next error.  I'm reworking this a little in another PR, so not worrying about improving it now.

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+export function getFriendlyPathToParamFile(templateUri: Uri, paramFileUri: Uri): string {
+  const workspaceFolder = workspace.getWorkspaceFolder(paramFileUri);
+  if (workspaceFolder) {
+    return path.relative(path.dirname(templateUri.fsPath), paramFileUri.fsPath);
+  } else {
+    return paramFileUri.fsPath;
+  }
+}
+
+interface QuickPickList {
```
alexgav: >QuickPickList [](start = 10, length = 13)

Nit: should interface name start with an I

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+
+  // Create initial pick list and sort it
+  let pickItems: IAzureQuickPickItem<IPossibleParamFile | undefined>[] = suggestions.map(paramFile => createQuickPickItem(paramFile, currentParamFile, templateUri));
+  sortQuickPickList(pickItems);
+
+  // Move the current item (if any) to the top
+  const currentItem = pickItems.find(pi => pi.data === currentParamFile);
+  if (currentItem) {
+    // tslint:disable-next-line: no-any
+    pickItems = [currentItem].concat(pickItems.filter(ppf => ppf !== currentItem));
```
alexgav: >pickItems [](start = 4, length = 9)

Nit: Can't we just use array indexes to swap instead of rebuilding the array?

src/AzureRMTools.ts
StephenWeatherford: Believe I got all functions, types etc, but didn't worry about local vars.

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+export function findMappedParamFileForTemplate(templateFileUri: Uri): Uri | undefined {
+  const paramFiles: { [key: string]: string } | undefined =
+    workspace.getConfiguration(configPrefix).get<{ [key: string]: string }>(configKeys.parameterFiles)
+    // tslint:disable-next-line: strict-boolean-expressions
+    || {};
+  if (typeof paramFiles === 'object') {
+    const normalizedTemplatePath = normalizePath(templateFileUri.fsPath);
+    let paramFile: Uri | undefined;
+    for (let fileNameKey of Object.getOwnPropertyNames(paramFiles)) {
+      const normalizedFileName: string | undefined = normalizePath(fileNameKey);
```
alexgav: >normalizePath [](start = 53, length = 13)

Don't we save them normalized? Do we need to do this again? Also, do we need the loop - can we just lookup property value via property name [propertyName]?

src/parameterFiles.ts
StephenWeatherford: I didn't worry about it here because it's this instance only and likely to match. Also, consequences of being wrong are minor.  Will add comments.

src/parameterFiles.ts
alexgav: Ah, nevermind, looks like we don't save them normalized. Makes sense I guess since that's what the user will be seeing. Even on Windows they would want case to match to what's in the filesystem.

---
In reply to: [389204136](https://github.com/microsoft/vscode-azurearmtools/pull/514#discussion_r389204136) [](ancestors = 389204136)

src/parameterFiles.ts
StephenWeatherford: DeploymentTemplate needs to full file. Also, user opened it, so it's not a concern like when reading a bunch of random files of who knows what length and that may not be intended for this use.

src/parameterFiles.ts
StephenWeatherford: Yeah, but this is easy to understand, less likely to get wrong, and performance not an issue.

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+
+async function doesFileContainString(filePath: string, matches: (fileSubcontents: string) => boolean, maxBytesToRead: number): Promise<boolean> {
+  // tslint:disable-next-line: typedef
+  return new Promise<boolean>((resolve, reject) => {
+    const stream = fse.createReadStream(filePath, { encoding: 'utf8' });
+
+    let content: string = '';
+    stream.on('data', (chunk: string) => {
+      content += chunk;
+      if (matches(content)) {
```
alexgav: >matches [](start = 10, length = 7)

Nit: probably could look just in the chunk + length of desired match before that rather than looking in the entire content on every chunk.

src/parameterFiles.ts
StephenWeatherford: No, this is from user settings and thus modifiable by the user.  Can't do a simple lookup because the case might be different.

src/parameterFiles.ts
```diff
@@ -0,0 +1,507 @@
...
+ */
+export async function findSuggestedParameterFiles(templateUri: Uri): Promise<IPossibleParamFile[]> {
+  let paths: IPossibleParamFile[] = [];
+
+  // Current logic is simple: Find all .json/c files in the same folder as the template file and check
+  //   if they're a parameter file
+  try {
+    const folder = path.dirname(templateUri.fsPath);
+    const fileNames: string[] = await fse.readdir(folder);
+    for (let paramFileName of fileNames) {
```
alexgav: >for (let paramFileName of fileNames)  [](start = 4, length = 37)

Would it make sense to log in telemetry length of time it took us to search all files? Mostly concerned that some people (and they will be those users :) ) may have thousands of files in one folder.

alexgav: And if we did see long delays, then maybe be more selective about naming patterns. Aren't they typically .parameters.json? 

---
In reply to: [389206328](https://github.com/microsoft/vscode-azurearmtools/pull/514#discussion_r389206328) [](ancestors = 389206328)

src/parameterFiles.ts
StephenWeatherford: True, but can't just check the last chunk, it might be split between chunks. So we'd have to have knowledge about the length of the string to check, which bleeds implementation details from schema.ts.  I think probably not worth worrying about.

src/parameterFiles.ts
StephenWeatherford: Not a bad idea.  Common patterns are .params and .parameters, but the could be anything.

StephenWeatherford: Wrapped with callWithTelemetryAndErrorHandling, which will automatically add timing.

StephenWeatherford:
Just FYI, we've had some issues with the Azure schema CDN and some other things that have delayed this release.  Will have it out as soon as possible, hopefully next week.

PR #522, opened 2020-03-06
ARM snippet updates (S - Z)

neilpeterson:
- Updated API versions to latest known
- Updated snippets so that they pass validation
- Updated storage account to `2019-06-01` which includes schema updates (also completed on VM snippets)
- Updated VMs to look more like [this doc examples](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/ps-template). This included removing the diagnostic extension from the windows VM, and adding a network security group to both. Both VMs snippets have been deployed to validate functionality.


assets/armsnippets.jsonc
```diff
@@ -1818,14 +1820,16 @@
             "{",
             "    \"name\": \"${1:storageaccount1}\",",
             "    \"type\": \"Microsoft.Storage/storageAccounts\",",
-            "    \"apiVersion\": \"2015-06-15\",",
-            "    \"location\": \"[resourceGroup().location]\",",
+            "    \"apiVersion\": \"2019-06-01\",",
```
StephenWeatherford: I'd stick to the property ordering suggested in https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#best-practices, except perhaps for name, which we've consistently got first?  Maybe we should change those eventually?

assets/armsnippets.jsonc
```diff
@@ -1884,19 +1888,20 @@
...
             "    \"tags\": {",
             "        \"displayName\": \"${1:ubuntuVM1} Storage Account\"",
             "    },",
-            "    \"properties\": {",
-            "        \"accountType\": \"Standard_LRS\"",
-            "    }",
+            "    \"sku\": {",
+            "       \"name\": \"${2:Standard_LRS}\"",
+            "    },",
+            "    \"kind\": \"${3|Storage,StorageV2,BlobStorage,BlockBlobStorage,FileStorage|}\"",
```
StephenWeatherford: I would think not all of these would be appropriate for use as a storage account for a VM.  My guess would be StorageV2 would be recommended, but you should see if there's a standard one recommended.

assets/armsnippets.jsonc
neilpeterson: I will create a work item to scrub / update the complete tempalte file to align with the best practices doc.

assets/armsnippets.jsonc
```diff
@@ -2225,19 +2265,20 @@
...
             "    \"tags\": {",
             "        \"displayName\": \"${1:windowsVM1} Storage Account\"",
             "    },",
-            "    \"properties\": {",
-            "        \"accountType\": \"Standard_LRS\"",
-            "    }",
+            "    \"sku\": {",
+            "       \"name\": \"${2:Standard_LRS}\"",
+            "    },",
+            "    \"kind\": \"${3|Storage,StorageV2,BlobStorage,BlockBlobStorage,FileStorage|}\"",
```
StephenWeatherford: ditto

assets/armsnippets.jsonc
neilpeterson: Yes, good point. Looking into this now.

assets/armsnippets.jsonc
neilpeterson: Fixed / aligned with portal

assets/armsnippets.jsonc
neilpeterson: Fixed / aligned with portal

assets/armsnippets.jsonc
neilpeterson: https://github.com/microsoft/vscode-azurearmtools/issues/525

PR #551, opened 2020-03-24
Don't insert parens for completions (0.10.0)

### LINKED ISSUE #361: variable autocomplete does not show variables without quotes
When autocompleting a variable, the function name will autocomplete with:

variables()

and the cursor is between the parens...  If you start typing, the list of variables does not appear, but a generic list.  You have to type a single quote before you get the variable list...  Seems like it would make more sense to show the variable list without requiring the single quote? That will be the common case, generic list is probably never helpful here.

### LINKED ISSUE #501: Suggestion: Auto-complete should not insert ()
Experience in TS:

1. Find an item to auto-complete
   image
2. Press tab:
   image
3. Press `(`:
   image
4. Start typing var name:
   image

---

Experience with ARM tools:

1. Find an item to auto-complete:
   image
2. Press tab:
   image
3. Press `'`:
   image

Notice signature help doesn't show up, this is particularly bad since very few users actually know the ctrl+shift+space shortcut. Instead of inserting the brackets the extension should be relying on VS Code's core capability to auto-insert closing brackets, this also has the benefits of respecting settings like this (which I happen to use as well):

```
"editor.autoClosingBrackets": "never"
```

Ideally this would be the state when `'` is pressed:

image

StephenWeatherford:
Fixes #501
Fixes #361 
Related to #556

@neilpeterson This is a quick change that these things:
1) stops inserting open parens for function calls
2) stops inserting "." when completing a user-defined namespace in a function call
3) adds "(" and "," to Intellisense triggers

CONSIDER: How to keep completion list showing after typing a space (e.g. "concat(1,<space>")

I think this really improves the flow.  
It also fixes this bug: https://github.com/microsoft/vscode-azurearmtools/issues/361

### DIFF
--- a/src/AzureRMTools.ts
+++ b/src/AzureRMTools.ts
@@ -485,7 +485,8 @@ export class AzureRMTools {
                 return this.onProvideCompletionItems(document, position, token);
             }
         };
-        ext.context.subscriptions.push(vscode.languages.registerCompletionItemProvider(armDeploymentDocumentSelector, completionProvider, "'", "[", "."));
+        ext.context.subscriptions.push(vscode.languages.registerCompletionItemProvider(
+            armDeploymentDocumentSelector, completionProvider, "'", "[", ".", "("));
 
         const definitionProvider: vscode.DefinitionProvider = {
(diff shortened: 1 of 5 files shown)


neilpeterson:
Awesome, thanks. Timely, I was just looking at https://github.com/microsoft/vscode-azurearmtools/issues/501.

PR #562, opened 2020-03-30
Inserting 20177.9, using zip for schemas

StephenWeatherford:


package.json
```diff
@@ -5,7 +5,7 @@
     "version": "0.9.0-alpha",
     "publisher": "msazurermtools",
     "config": {
-        "ARM_LANGUAGE_SERVER_NUGET_VERSION": "3.0.0-preview.20175.11"
+        "ARM_LANGUAGE_SERVER_NUGET_VERSION": "3.0.0-preview.20177.9"
```
alexgav: We should use 3.0.0-preview.20180.1

PR #604, opened 2020-04-16
[0.9.1] Use new .net tooling extension for dotnet acquisition

StephenWeatherford:


StephenWeatherford:
Suites aren't passing yet due to needing to install the .net extension during tests

Issue #598, opened 2020-04-14
This extension is trying to install a legacy dot net core version


j03INTHECLOUD:
Error:
Downloading .NET Core tooling version(s) 3.0.2........[dotnet 3.0.2]: dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Runtime/3.0.2/dotnet-runtime-3.0.2-win-x64.zip
dotnet-install: Cannot download: https://dotnetcli.azureedge.net/dotnet/Runtime/3.0.2/dotnet-runtime-3.0.2-win-x64.zip
dotnet-install: Downloading legacy link: https://dotnetcli.azureedge.net/dotnet/Runtime/3.0.2/dotnet-win-x64.3.0.2.zip
dotnet-install: Cannot download: https://dotnetcli.azureedge.net/dotnet/Runtime/3.0.2/dotnet-win-x64.3.0.2.zip

cox-antohi:
My solution was to to add 
`[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12`
inside
**C:\Users\USERNAME\\.vscode\extensions\msazurermtools.azurerm-vscode-tools-0.9.0\assets\install scripts\dotnet-install.ps1**
after the line with
`$HttpClient.Timeout = New-TimeSpan -Minutes 20`

It seems to me that the next line having the `$HttpClient.GetAsync` command is getting **Could not create SSL/TLS secure channel.**.

robmatear:
> My solution was to to add
> `[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12`
> inside
> **C:\Users\USERNAME\.vscode\extensions\msazurermtools.azurerm-vscode-tools-0.9.0\assets\install scripts\dotnet-install.ps1**
> after the line with
> `$HttpClient.Timeout = New-TimeSpan -Minutes 20`
> 
> It seems to me that the next line having the `$HttpClient.GetAsync` command is getting **Could not create SSL/TLS secure channel.**.

This solution worked for me (it installed 3.0.2) but there is still the problem of the script targeting an old version of the runtime when a newer one is available.

StephenWeatherford:
@j03INTHECLOUD @cox-antohi @robmatear Thanks for reporting!

We're about to ship a hot fix that changes the way we retrieve .net core, and should avoid this problem in the future.

StephenWeatherford:
[...]

PR #616, opened 2020-04-17
Use dotnet 3.0.3, support TLS 2/3

StephenWeatherford:
The first commit updates the install scripts to the latest version, although it's mostly whitespace changes.  It also uses 3.0.3 instead of 3.0.2
The second commit attempts to add the TLS 2/3 fix from https://github.com/microsoft/vscode-azurearmtools/issues/598#issuecomment-614855482 and https://github.com/microsoft/vscode-azurearmtools/issues/598#issuecomment-614708618

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the "vscode-azurearmtools" extension will be documented in this file.
 
+## Version 0.9.1 (2020-04-17)
+
+### Fixed
+
+- Update to 3.0.3 of .NET Core runtime and update install scripts
+
 ## Version 0.9.0 (2020-04-14)
 
 ### Added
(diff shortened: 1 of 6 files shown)


sfoslund:
Adding @bozturkMSFT and @donJoseLuis as reviewers, as they own the install script. 

I don't know enough in this area to know if this is a safe change to make given [similar changes to the install script](https://github.com/dotnet/website/pull/1877) have been backed out.

bozturkMSFT:
> Edit: I got come more context on this issue and it sounds like it's safe to set TLS before the script, but not inside of it. I have a draft PR for this change in [dotnet/vscode-dotnet-runtime#60](https://github.com/dotnet/vscode-dotnet-runtime/pull/60)

@sfoslund Can you give more context on this? Why is it safe to set it before but not inside?

sfoslund:
@mmitche can you give an explanation on setting TLS before invoking the script vs during?

StephenWeatherford:
Superseded.

PR #620, opened 2020-04-18
Support TLS12/13

StephenWeatherford:
The problem as I understand it is that machines with .NET 4.5 don't have TLS1.2 enabled by default (and something changed in Azure enforcing certain security requirements).


enzanglach:
> 
> 
> The problem as I understand it is that machines with .NET 4.5 don't have TLS1.2 enabled by default (and something changed in Azure enforcing certain security requirements).

Might be, it worked fine on my other Win10 machine where TLS1.2 have been enabled.

Issue #525, opened 2020-03-06
Ensure resources definitions align with best practices doc


neilpeterson:
Update ordering of properties etc..

https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#best-practices

PR #628, opened 2020-04-19
Snippets common property order

neilpeterson:
- Reorder common properties to match [best practices doc](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#best-practices)
- Removed empty tags
- Few small resource configuration updates

Fixes https://github.com/microsoft/vscode-azurearmtools/issues/525


StephenWeatherford:
How important is it that this order match the sorting template order?  Ideally these would both have options to set.

neilpeterson:
Thanks @StephenWeatherford. As discussed, this is just an update of the ordering for resource properties for snippets (type, apiVersion, name, etc.) which are not affected by the sorting functionality.

Issue #636, opened 2020-04-22
dateTimeAdd function not recognized


bergmeister:
`dateTimeAdd` is a supported function in ARM, see here: https://docs.microsoft.com/azure/azure-resource-manager/templates/template-functions-date#datetimeadd
However, the extension does not recognize it. Using version `0.9.1` of the extension.

![image](https://user-images.githubusercontent.com/9250262/79968093-95fc1b00-8487-11ea-99c6-84183dc74dd9.png)

alex-frankel:
FYI @neilpeterson, this is a brand new function. Here is the doc page: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-date#datetimeadd

neilpeterson:
Thanks @bergmeister and @alex-frankel I've added the function and pending no issues we will ship this with our next release.

https://github.com/microsoft/vscode-azurearmtools/pull/638

bergmeister:
Awesome, thanks

StephenWeatherford:
Shipped yesterday

PR #638, opened 2020-04-22
Date time function

neilpeterson:
https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-date#datetimeadd

Fixes https://github.com/microsoft/vscode-azurearmtools/issues/636


assets/ExpressionMetadata.json
```diff
@@ -99,6 +99,13 @@
             "minimumArguments": 1,
             "maximumArguments": 1
         },
+        {
+            "name": "dateTimeAdd",
+            "expectedUsage": "dateTimeAdd(base, duration, [format])",
+            "description": "Adds a time duration to a base datetime value.",
```
bergmeister, line 105: The type is actually string and not datetime
```suggestion
            "description": "Adds a time duration to a base value. ISO 8601 format is expected.",
```

neilpeterson: @alex-frankel can you advise? I've taken the description verbatim from docs - thanks.

https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-date#datetimeadd

alex-frankel: Yeah I guess datetime could be read as a type, when the intention was a format for the string. This looks good. @tfitzmac FYI - we should probably update the doc as well

tfitzmac: Thanks, I'll update the doc today.

StephenWeatherford:
@neilpeterson Not sure if the above issues are resolved - merge whenever you're ready.

bergmeister:
Awesome, when do you plan to release this change?

neilpeterson:
This update will be in 0.10.0, rough release estimates are here:

https://github.com/microsoft/vscode-azurearmtools/milestones

PR #639, opened 2020-04-22
Removed extra comma and collapsed root template sections

neilpeterson:
Sorry @StephenWeatherford I missed this when looking at these changes. When the language mode is Azure Resource Manager Tempalte, the snippet updates include an extra comma after outputs. I've also updated to collapse these sections.
Before:

image

After:

image

Thanks


StephenWeatherford:
The reason I expanded them is that our Format Document collapses them...  So for consistency, but also because in my tests I'm using format document to make sure everything is formatted correctly in the snippet (i.e., if format document doesn't make any changes, I know the snippet was formatted correctly).  Thoughts?

StephenWeatherford:
It might be possible to modify our format document behavior if you really don't like it, not sure.  There were some settings under our control.

StephenWeatherford:
Whoops, didn't mean to merge that one, oh well.

neilpeterson:
I will add a work item and get to it asap.

PR #645, opened 2020-04-24
Improve Intellisense experience for new completions work

### LINKED ISSUE #626: Improve the feel of auto-completion for resourceId, parameters
- [x] filtering
- [x] resourceId
  - [x] filtering
  - [x] partial completions
- [x] Partial completions
  - [x] "&lt;new parameter&gt;"
  - [x] "p1"
  - [x] typing "," on end of line brings up intellisense - change this?
image

  - [x] "Any character triggers it"
  - [x] Leaves around the previous word

StephenWeatherford:
Fixes #626 

The main part of the fix is to make sure the replacement span includes the full area that should be used for filtering and that will be replaced when a completion is selected.


StephenWeatherford:
> Perhaps this not exactly related, should be a separate work item, or out of scope, but I am noticing that `ctrl + space` does not present the new parameter completion when in the correct cursor position.

Fixed

PR #698, opened 2020-05-11
Pick up new schema zip

StephenWeatherford:


StephenWeatherford:
@neilpeterson Worked around it in the suites. If you want to change the apiVersion in the snippet to avoid the error, I'll leave that to you.

Issue #700, opened 2020-05-12
Microsoft.Logic/workflows snippet apiVersion references removed schema


neilpeterson:
The Microsoft.Logic/workflows schema version 2019-05-01 is valid, however the schema has been removed from the schemas repository. I've opened an issue to track.

https://github.com/Azure/azure-resource-manager-schemas/issues/1007

We should update the Logic App snippet to use a version with existing schema (2017-07-01).

PR #701, opened 2020-05-12
Updated API version 2019-05-01 > 2017-07-01

neilpeterson:
Fixes https://github.com/microsoft/vscode-azurearmtools/issues/700


assets/armsnippets.jsonc
```diff
@@ -1715,7 +1715,7 @@
             "{",
             "\t\"name\": \"${1:logicApp1}\",",
             "\t\"type\": \"Microsoft.Logic/workflows\",",
-            "\t\"apiVersion\": \"2019-05-01\",",
+            "\t\"apiVersion\": \"2017-07-01\",",
```
StephenWeatherford, line 1718: This will cause my work-around to fail.  Could you run the "Logic App" suite (via launch.json) and make sure it's fixed?  Thanks!

PR #720, opened 2020-05-20
added listKeyValue TLE for app config

bmoore-msft:
came up in a build session, not documented for templates yet but adding it here.  See:

https://docs.microsoft.com/en-us/rest/api/appconfiguration/configurationstores/listkeyvalue


assets/ExpressionMetadata.json
```diff
@@ -458,6 +458,18 @@
                 }
             ]
         },
+        {
+            "name": "listKeyValue",
+            "expectedUsage": "listKeyValue(resourceName\/resourceIdentifier, apiVersion, KeyLabelObject)",
```
StephenWeatherford, line 463: nit: 
```suggestion
            "expectedUsage": "listKeyValue(resourceName\/resourceIdentifier, apiVersion, keyLabelObject)",
```

assets/ExpressionMetadata.json
```diff
@@ -458,6 +458,18 @@
...
+            "description": "Returns a value from the App Config store with Key/Label specified in the request. The KeyLabelObject parameters is a JSON object with the two properties \"key\" and \"label\" as required by the values in the store.",
```
StephenWeatherford, line 464: How about an example?  Some of these are difficult to figure out.  For instance, I'm guessing keyLabelObject has to be a variable/param reference?  Is there a way to create a JSON object directly in TLE?

assets/ExpressionMetadata.json
```diff
@@ -458,6 +458,18 @@
...
+            "minimumArguments": 3,
+            "maximumArguments": 3,
+            "returnValueMembers": [
+                {
+                    "name": "value"
+                }
+            ]
+        },
         {
```
StephenWeatherford, line 473: We could also relatively easily add a field for a documentation link (https://docs.microsoft.com/en-us/rest/api/appconfiguration/configurationstores/listkeyvalue), if you think that makes sense.

assets/ExpressionMetadata.json
bmoore-msft: Can we provide json within the json?  Can certainly do that...

assets/ExpressionMetadata.json
bmoore-msft: It's not doc'd yet (separate bug for that) but we could link here: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions#resource-functions

We don't actually doc each function in the template docs - could reference the REST API though that's a little out of context.  Could also reference quickstart but that link would eventually get broken

assets/ExpressionMetadata.json
StephenWeatherford: So, did you want to add something to the description here to clarify usage or leave it alone?

bmoore-msft: I'm fine with it as is...

Issue #484, opened 2020-02-11
Variables in nested templates fails validation


JakobGSvendsen:
I am making a nested template. with the template content in the same file (so nested, not linked).
and using "inner" scope as described here:

https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/linked-templates#scope-for-expressions-in-nested-templates

I am using variables inside the nested template, but the tools reports error while there is none:
![image](https://user-images.githubusercontent.com/3460807/74234383-fa06a400-4ccc-11ea-9c9a-22643daa7da2.png)

It seems that this extesnion does not support scope in nested templates.

Templates works fine, it is only in the editor that it reports errors as it only checks the main variables not the nested ones

neilpeterson:
Thanks @JakobGSvendsen we are aware of the issue and will update here once the work has been completed.

https://devdiv.visualstudio.com/DevDiv/_boards/board/t/ARM%20Template%20Authoring/Stories/?workitem=824392

neilpeterson:
Example that can be used for validation.

https://github.com/neilpeterson/arm-template-collection/blob/master/1-priority-work-examples/nested-example/azuredeploy.json

StephenWeatherford:
# Development Notes and Issues
- [ ] resourceId completion should work within the current scope

holylander:
Thanks! I also found this issue. Looking forward to the fix :)

StephenWeatherford:
You shouldn't have to wait long...  :-)

StephenWeatherford:
Fix released with 0.11.0

PR #722, opened 2020-05-21
Support for nested templates, part 1

StephenWeatherford:
Most scenarios for https://github.com/microsoft/vscode-azurearmtools/issues/484.  See https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/linked-templates#expression-evaluation-scope-in-nested-templates - nested templates can either use the parent scope ("outer") or their own scope ("inner").  We don't currently support inner.

- Refactored TemplateScope to class hierarchy
- Moved resources to TemplateScope since nested templates have their own resources
- Rewrote quotedStringToTleParseResultMap to handle properly assigning scopes to the tree, including multiply-nested scopes.
- Moved code to get params/vars/etc from DeploymentTemplate to a TemplateScope subclass.
- Added Resource/IResource to more easily handle the fact that nested templates are defined inside deployment-typed resources.
- Tests
- Enable nested-deployment-scoping.json/azuredeploy.inline.json tests which were blocked by this.


src/DeploymentTemplate.ts
```diff
@@ -367,112 +328,6 @@ export class DeploymentTemplate extends DeploymentDocument {
         return count;
     }
 
-    private getTopLevelParameterDefinitions(): ParameterDefinition[] {
```
StephenWeatherford, line 370: These were simply moved to TemplateScopeFromObject

src/Resource.ts
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Convenient way of seeing what this object represents in the debugger, shouldn't be used for production code
+     */
+    public get __debugDisplay(): string {
+        return this.nameValue?.unquotedValue ?? '(unnamed resource)';
+    }
+
+    public get childDeployment(): TemplateScope | undefined {
+        return this._childTemplate.getOrCacheValue(() => {
+            return getChildTemplateForResourceObject(this.parentScope, this.resourceObject);
```
alexgav, line 34: >getChildTemplateForResourceObject [](start = 19, length = 33)

Would it make sense for this to be a method of the Resource class rather than an external utility method? Or does it get called with values that are not fields of this object?

src/TemplateScope.ts
```diff
@@ -3,50 +3,96 @@
...
  * Represents the scoped access of parameters/variables/functions at a particular point in the template tree.
  */
-export class TemplateScope {
-    /**
-     * Constructor
-     */
+export abstract class TemplateScope {
+    private _parameterDefinitions: CachedValue<IParameterDefinition[] | undefined> = new CachedValue<IParameterDefinition[] | undefined>();
+    private _variableDefinitions: CachedValue<IVariableDefinition[] | undefined> = new CachedValue<IVariableDefinition[] | undefined>();
+    private _functionDefinitions: CachedValue<UserFunctionNamespaceDefinition[] | undefined> = new CachedValue<UserFunctionNamespaceDefinition[] | undefined>();
```
alexgav, line 30: >UserFunctionNamespaceDefinition [](start = 46, length = 31)

Nit: I don't fully know the codebase, but it looks a bit weird for UserFunctionNameSpaceDefinition to be accessed via class rather than interface as all other xxxDefinitions, no?

src/TemplateScope.ts
```diff
@@ -3,50 +3,96 @@
...
+    private _resources: CachedValue<IResource[] | undefined> = new CachedValue<IResource[] | undefined>();
+
+    public readonly abstract scopeKind: TemplateScopeKind;
```
alexgav, line 33: >TemplateScopeKind [](start = 40, length = 17)

Do all fields (all definition kinds) make sense for all TemplateScopeKind values?

src/TemplateScope.ts
```diff
@@ -3,50 +3,96 @@
...
+            ?? [];
     }
 
     public get namespaceDefinitions(): UserFunctionNamespaceDefinition[] {
-        // tslint:disable-next-line: strict-boolean-expressions
-        return this._namespaceDefinitions || [];
+        return this._functionDefinitions.getOrCacheValue(() => this.getNamespaceDefinitions())
+            ?? [];
+
+    }
```
alexgav, line 74: Nit: empty line

src/templateScopes.ts
```diff
@@ -0,0 +1,347 @@
...
+        }
+    }
+
+    return [];
+}
+
+function getNamespaceDefinitionsFromObject(objectValue: Json.ObjectValue | undefined): UserFunctionNamespaceDefinition[] {
+    const namespaceDefinitions: UserFunctionNamespaceDefinition[] = [];
+
+    // Example of function definitions
```
alexgav, line 147: > Example of function definitions [](start = 6, length = 32)

Thanks you for these, very helpful.

src/templateScopes.ts
```diff
@@ -0,0 +1,347 @@
...
+        ?.unquotedValue;
+    if (resourceTypeLC?.toLowerCase() === deploymentsResourceTypeLC) {
+        // Get the template property under properties
+        const nestedTemplateObject =
+            resourceObject?.getPropertyValue(templateKeys.properties)?.asObjectValue
+                ?.getPropertyValue('template')?.asObjectValue;
+        if (nestedTemplateObject) {
+            const scopeKind = getExpressionScopeKind(resourceObject);
+            const templateName: string = resourceObject?.getPropertyValue(templateKeys.resourceName)?.asStringValue?.unquotedValue
+                ?? '(unnamed template';
```
alexgav, line 318: >' [](start = 37, length = 1)

Nit: missing close paren

src/templateScopes.ts
```diff
@@ -0,0 +1,347 @@
...
+    //                    ..
+
+    // Correct resource type?
+    const resourceTypeLC = resourceObject?.getPropertyValue(templateKeys.resourceType)?.asStringValue
+        ?.unquotedValue;
+    if (resourceTypeLC?.toLowerCase() === deploymentsResourceTypeLC) {
+        // Get the template property under properties
+        const nestedTemplateObject =
+            resourceObject?.getPropertyValue(templateKeys.properties)?.asObjectValue
+                ?.getPropertyValue('template')?.asObjectValue;
```
alexgav, line 314: >template [](start = 36, length = 8)

Nit: create a constant in templateKeys?

src/templateScopes.ts
```diff
@@ -0,0 +1,347 @@
...
+    //             "expressionEvaluationOptions": {
+    //                 "scope": "inner"
+    //             },
+    //             "template": {
+    //                 "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    //                 "contentVersion": "1.0.0.0",
+    //                 "parameters": {
+    //                    ..
+
+    // Correct resource type?
```
alexgav, line 307: >/ Correct resource type? [](start = 5, length = 24)

Can't you use your isResourceADeployment helper here?

src/templateScopes.ts
```diff
@@ -0,0 +1,347 @@
...
+
+export function isResourceADeployment(resourceObject: Json.ObjectValue): boolean {
+    const resourceTypeLC = resourceObject?.getPropertyValue(templateKeys.resourceType)?.asStringValue
+        ?.unquotedValue;
+    return resourceTypeLC?.toLowerCase() === deploymentsResourceTypeLC;
+}
+
+function getExpressionScopeKind(resourceObject: Json.ObjectValue | undefined): ExpressionScopeKind {
+    const scopeValue = resourceObject?.getPropertyValue(templateKeys.properties)?.asObjectValue
+        ?.getPropertyValue('expressionEvaluationOptions')?.asObjectValue
```
alexgav, line 341: >expressionEvaluationOptions [](start = 28, length = 27)

Nit: add to templateKeys? Same with scope?

src/templateScopes.ts
```diff
@@ -0,0 +1,347 @@
...
+        ?.getPropertyValue('scope')
+        ?.asStringValue?.unquotedValue;
+    return scopeValue?.toLowerCase() === 'inner'
```
alexgav, line 344: >'inner' [](start = 41, length = 7)

Nit: compare with ExpressionScopeKind.inner?

src/UserFunctionDefinition.ts
```diff
@@ -45,24 +46,13 @@ export class UserFunctionDefinition implements INamedDefinition {
     public get scope(): TemplateScope {
         return this._scope.getOrCacheValue(() => {
             // Each user function has a scope of its own
-            return new TemplateScope(
-                ScopeContext.UserFunction,
-                // User functions can only use their own parameters, they do
-                //   not have access to top-level parameters
+            return new UserFunctionScope(
```
alexgav, line 49: >UserFunctionScope [](start = 23, length = 17)

Nice, much cleaner

src/DeploymentTemplate.ts
```diff
@@ -542,3 +397,53 @@ export class DeploymentTemplate extends DeploymentDocument {
...
+    private readonly _scopeStack: TemplateScope[] = [];
+    private _currentScope: TemplateScope;
+    private readonly _allScopesInTemplate: TemplateScope[] = [];
+
+    public constructor(private readonly _dt: DeploymentTemplate) {
+        super();
+        this._currentScope = _dt.topLevelScope;
+    }
+
+    public createParsedStringMap(): Map<Json.StringValue, TLE.ParseResult> {
```
alexgav, line 412: >createParsedStringMap [](start = 11, length = 21)

This is a little weird to me in that if you call it multiple times on the same object, it will keep modifying the object, so you'll get different results, right?

src/TemplateScope.ts
StephenWeatherford: Considered changing that (does look weird), but decided there wasn't really any advantage to doing it. For Typescript, it's a lot easer to switch to an interface when it's actually needed, so would rather keep things simple until the extra abstraction's actually needed.

src/Resource.ts
StephenWeatherford: It would, in fact I tried to do that, but that causes a circular dependence Resource -> TemplateScope -> Resource.

src/templateScopes.ts
StephenWeatherford: Had inlined it, forgot to get rid of isResourceADeployment, done.

src/templateScopes.ts
StephenWeatherford: Created new constant.

src/TemplateScope.ts
StephenWeatherford: Need to be available to be queried, yes (some may be unsupported)

PR #726, opened 2020-05-23
Adding in Resource Group snippet

kilasuit:
Adds in a snippet for resource groups for Subscription level deployments


assets/armsnippets.jsonc
```diff
@@ -1940,6 +1940,24 @@
...
+        "prefix": "arm-rg",
+        "body": [
+            "{",
+            "\t\"name\": \"${1:resourceGroup}\",",
+            "\t\"type\": \"Microsoft.Resources/resourceGroups\",",
+            "\t\"apiVersion\": \"2019-10-01\",",
+            "\t\"location\": \"[parameters('location')]\",",
+            "\t\"dependsOn\": [",
+            "\t],",
+            "\t\"tags\": [",
```
neilpeterson, line 1953: Tags should be an object vs. array. Change the square brackets to curly braces.

https://docs.microsoft.com/en-us/azure/templates/microsoft.resources/2019-10-01/resourcegroups#microsoftresourcesresourcegroups-object

assets/armsnippets.jsonc
```diff
@@ -1940,6 +1940,24 @@
...
         "description": "Public IP Prefix"
     },
+    "Resource Group": {
+        "prefix": "arm-rg",
+        "body": [
+            "{",
+            "\t\"name\": \"${1:resourceGroup}\",",
+            "\t\"type\": \"Microsoft.Resources/resourceGroups\",",
+            "\t\"apiVersion\": \"2019-10-01\",",
+            "\t\"location\": \"[parameters('location')]\",",
```
neilpeterson, line 1950: Consider using a tab stop for location. Because a parameter with the name location may not exist, this would give the consume a out of the box utility for updating the value. The update would look like this:

```
"\t\"location\": \"${2:[parameters('location')]}\",",
```

assets/armsnippets.jsonc
```diff
@@ -1940,6 +1940,24 @@
...
+            "\t\"dependsOn\": [",
+            "\t],",
+            "\t\"tags\": [",
+            "\t],",
+            "\t\"properties\": {",
```
neilpeterson, line 1955: nit for consideration; properties are not necessary here and can be removed for brevity.

neilpeterson:
@kilasuit Thanks a ton for this update, really appreciate it. For the most part looks good, I've added a few things to update and / or consider. 

If you have not seen this, snippets can be tested by debugging the project. To do so, navigate to the debug view, and click the green triangle. This will open a new instance of VS Code with the updated snippets available for use.

image

neilpeterson:
@kilasuit apologies for the delay on this. This PR will require an update to our test suites which I have submitted here - https://github.com/microsoft/vscode-azurearmtools/pull/756. Once this makes it into master we can proceed with the snippet update.

One last thing will need to be updated. Instead of using a parameter as the location value, can you enter a location value, with a tab stop? The update will look like this.

```
"\t\"location\": \"${2:eastus}\",",
```

Thanks

StephenWeatherford:
@kilasuit Just FYI, if you want this in the next release, we'll need the updates in the next week days.  No rush, though, it can always go in later.

PR #742, opened 2020-06-01
Code lens for parameters and nested templates

### LINKED ISSUE #675: Add code lenses for parameters
1. At the "parameters" section to point to the selected parameter file
1. At parameter definitions to indicate their effective value

StephenWeatherford:
Fixes part of https://github.com/microsoft/vscode-azurearmtools/issues/484
Fixes #675 

image
image

image
image


src/AzureRMTools.ts
```diff
@@ -188,13 +190,34 @@ export class AzureRMTools {
...
+                    //   of file.
+                    const span = dp.getParameterValue(param)?.value?.span
+                        ?? dp.parametersProperty?.nameValue.span
+                        ?? new Language.Span(0, 0);
+                    const range = getVSCodeRangeFromSpan(dp, span);
+                    editor.selection = new vscode.Selection(range.start, range.end);
+                    editor.revealRange(range);
+                }
+            });
+
```
alexgav, line 220: Nit: would it make sense to make this a class method rather than an anonymous function here? Seems like that would be the pattern the rest of the command handlers are following.

src/DeploymentTemplate.ts
```diff
@@ -395,17 +398,79 @@ export class DeploymentTemplate extends DeploymentDocument {
...
+        return allScopes;
+
+        function traverse(scope: TemplateScope | undefined): void {
+            for (let childScope of scope?.childScopes ?? []) {
+                if (allScopes.indexOf(childScope) < 0) {
+                    allScopes.push(childScope);
+                } else {
+                    let a = 1;
+                    a = a;
+                }
```
alexgav, line 454: Why are we doing this?

StephenWeatherford: I know got rid of that in some branch somewhere...  :-)

It was just a place to put a breakpoint for easy debugging.

src/AzureRMTools.ts
StephenWeatherford: Will do.  This whole file needs refactoring.

StephenWeatherford:
Thanks!

PR #743, opened 2020-06-02
Fix warnings for nested scopes and warn about inaccessible inner scope members

StephenWeatherford:
Fixes part of #484 

Two things:
1) We weren't looking for warnings in deeper scopes
2) Added new warning about inaccessible members for outer-scoped nested templates.

### DIFF
--- a/src/DeploymentTemplate.ts
+++ b/src/DeploymentTemplate.ts
@@ -255,25 +255,29 @@ export class DeploymentTemplate extends DeploymentDocument {
             + `If you intended inner scope, set properties.nestedDeploymentExprEvalOptions.scope to 'inner'.`;
 
         for (const scope of this.allScopes.filter(ts => ts.scopeKind === TemplateScopeKind.NestedDeploymentWithOuterScope)) {
-            const parameters = scope.rootObject?.getProperty(templateKeys.parameters)?.value;
+            const parameters = getPropertyValueOfScope(templateKeys.parameters);
             // tslint:disable-next-line: strict-boolean-expressions
             if (!!parameters?.asObjectValue?.properties?.length) {
                 warnings.push(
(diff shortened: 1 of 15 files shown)


src/DeploymentTemplate.ts
```diff
@@ -228,21 +243,37 @@ export class DeploymentTemplate extends DeploymentDocument {
...
+     */
+    private findInaccessibleScopeMembers(): language.Issue[] {
         const warnings: language.Issue[] = [];
+        const warningMessage =
+            // tslint:disable-next-line: prefer-template
+            'Variables, parameters and user functions of an outer-scoped nested template are inaccessible to any expressions. '
+            + `If you intended inner scope, set properties.nestedDeploymentExprEvalOptions.scope to 'inner'.`;
+
+        for (const scope of this.allScopes.filter(ts => ts.scopeKind === TemplateScopeKind.NestedDeploymentWithOuterScope)) {
+            const parameters = scope.rootObject?.getProperty(templateKeys.parameters)?.value;
```
alexgav, line 258: >scope.rootObject?.getProperty(templateKeys.parameters)?.value [](start = 31, length = 61)

Nit: I personally would've used a helper function for this fairly complicated expression since it's repeated three times with a different property name. Same for !!parameters?.xxxx bool expression below. Actually, it looks like even a large helper could've been created that parameterized 

             const parameters = scope.rootObject?.getProperty(templateKeys.parameters)?.value;
            // tslint:disable-next-line: strict-boolean-expressions
            if (!!parameters?.asObjectValue?.properties?.length) {
                warnings.push(
                    new language.Issue(parameters.span, warningMessage, language.IssueKind.inaccessibleNestedScopeMembers));
            }

StephenWeatherford: Refactored what I could.  The difference between objects and arrays made it limited.

PR #768, opened 2020-06-05
Handle UTF-8 BOM

### LINKED ISSUE #721: Fails to create a parameter file for arm template when encoded with UTF-8 BOM
Does this occur consistently? Yes 
Repro steps:

1. from vs code create a new arm template (use snippet arm! )
2. from vs code select/create parameter file...
I receive the error below

Action: azurerm-vscode-tools.selectParameterFile
Error type: Error
Error Message: "e:\temp\armtemp\mytemplate.json" does not appear to be an Azure Resource Manager deployment template file.

Version: 0.10.0
OS: win32
OS Release: 10.0.17763
Product: Visual Studio Code
Product Version: 1.45.1
Language: en

<details>
<summary>Call Stack</summary>

```
Object.<anonymous> extension.bundle.js:1:596601
fulfilled extension.bundle.js:1:593686
```

</details>

StephenWeatherford:
Fixes #721

### DIFF
--- a/src/util/readUtf8FileWithBom.ts
+++ b/src/util/readUtf8FileWithBom.ts
@@ -1,0 +1,15 @@
+// ----------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ----------------------------------------------------------------------------
+
+import * as fse from 'fs-extra';
+
+const utf8Bom = 65279;
+
+export async function readUtf8FileWithBom(fsPath: string): Promise<string> {
+    let contents = (await fse.readFile(fsPath, { encoding: "utf8" }));
+    if (contents.charCodeAt(0) === utf8Bom) {
+        contents = contents.slice(1);
+    }
+    return contents;
+}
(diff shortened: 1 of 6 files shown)


src/util/readUtf8FileWithBom.ts
```diff
@@ -0,0 +1,15 @@
...
+// ----------------------------------------------------------------------------
+
+import * as fse from 'fs-extra';
+
+const utf8Bom = 65279;
+
+export async function readUtf8FileWithBom(fsPath: string): Promise<string> {
+    let contents = (await fse.readFile(fsPath, { encoding: "utf8" }));
+    if (contents.charCodeAt(0) === utf8Bom) {
+        contents = contents.slice(1);
```
alexgav, line 12: >contents.slice(1) [](start = 19, length = 17)

Will this end up re-allocating the entire string/file contents? If yes, maybe would be more efficient to read the first char only first, and then read the rest of the file (starting at char 1 if char 0 is BOM?) Might possibly make a difference if someone is re-opening a workspace with a number of large templates previously opened.

StephenWeatherford: It's a good suggestion, but complicates the read logic, and the BOM should be a rare case (only a couple of users have hit this). And there's always GC.

PR #783, opened 2020-06-12
Fix keyvault values in code lens

StephenWeatherford:
Should have been looking for this format for keyvault references in parameter files:
```
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
      "adminLogin": {
        "value": "exampleadmin"
      },
      "adminPassword": {
        "reference": {  <<<<<<<<<<<<<<<<<<<<
          "keyVault": {
          "id": "/subscriptions/<subscription-id>/resourceGroups/<rg-name>/providers/Microsoft.KeyVault/vaults/<vault-name>"
          },
          "secretName": "ExamplePassword"
        }
      },
      "sqlServerName": {
        "value": "<your-server-name>"
      }
  }
}
```

But instead it was looking for this:
```
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
      "adminPassword": {
        "value": {     <<<<<<<<<<<<<<<<<<<<
          "reference": {
            "keyVault": {
            "id": "/subscriptions/<subscription-id>/resourceGroups/<rg-name>/providers/Microsoft.KeyVault/vaults/<vault-name>"
            },
            "secretName": "ExamplePassword"
          }
        }
      },
      "sqlServerName": {
        "value": "<your-server-name>"
      }
  }
}
```

### DIFF
--- a/src/deploymentTemplateCodeLenses.ts
+++ b/src/deploymentTemplateCodeLenses.ts
@@ -5,7 +5,7 @@
 // tslint:disable: max-classes-per-file
 
 import * as assert from 'assert';
-import { Json, Language } from '../extension.bundle';
+import { Language } from '../extension.bundle';
 import { DeploymentDocument, ResolvableCodeLens } from "./DeploymentDocument";
 import { DeploymentTemplate } from './DeploymentTemplate';
 import { IParameterDefinition } from './IParameterDefinition';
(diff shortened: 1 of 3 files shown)


src/deploymentTemplateCodeLenses.ts
```diff
@@ -78,18 +78,17 @@ export class ParameterDefinitionCodeLens extends ResolvableCodeLens {
...
             const defaultValueAsString = this.parameterDefinition.defaultValue?.toFullFriendlyString();
 
             let title;
-            if (givenValueAsString !== undefined) {
-                if (this.isKeyVaultReference(paramValue)) {
-                    title = 'Value: (KeyVault reference)';
-                } else {
-                    title = `Value: ${givenValueAsString}`;
-                }
+            if (!!paramReference) {
```
alexgav, line 88: >paramReference [](start = 18, length = 14)

I'm a little confused why we aren't looking for keyVault and id properties inside of a reference like we did before. Even if the structure changed, is it enough that there is a reference property in the parameter value? It would probably take me days to fully understand this, so I guess I'll trust you on this one. Doesn't seem like the code is doing what the PR description said it's doing.

StephenWeatherford: Don't know for sure, frankly.  But doesn't really seem to make sense to be overly picky about it.  It's definitely not a value, but also definitely not "value not found", which is what it would be otherwise.

Issue #746, opened 2020-06-02
Snippets: Put under our control (convert snippets to completions)


StephenWeatherford:
## TODO
- [x] Replace current word
- [x] Handle empty ARM file with arm! snippet
- [x] Shouldn't be adding backslash: "\$schema"
- [x] tab not working

PR #789, opened 2020-06-15
Convert snippet to completions

StephenWeatherford:
Fixes https://github.com/microsoft/vscode-azurearmtools/issues/746

This converts our snippets which are handled by vscode into our own logic which returns the snippets as completion items instead.  This will allow us to start being smarter about where the snippets are suggested, based on context.


StephenWeatherford:
@bhsubra Could you please review this?  Shouldn't be as hard as it looks, and I wouldn't worry too much about the test code. Thx.

package.json
```diff
@@ -75,10 +75,6 @@
             }
         ],
         "snippets": [
-            {
```
StephenWeatherford, line 78: Tells vscode to stop automically handling our ARM snippets.

src/insertItem.ts
```diff
@@ -43,43 +42,22 @@ export function getItemType(): QuickPickItem<ItemType>[] {
...
-    let snippetPath = path.join(assetsPath, "armsnippets.jsonc");
-    let content = fse.readFileSync(snippetPath, "utf8");
-    let tree = Json.parse(content);
-    if (!(tree.value instanceof Json.ObjectValue)) {
-        return items;
-    }
-    for (const property of tree.value.properties) {
-        if (isResourceSnippet(property)) {
-            items.push(getQuickPickItem(property.nameValue.unquotedValue));
+async function getResourceSnippets(): Promise<IAzureQuickPickItem<ISnippet>[]> {
```
StephenWeatherford, line 45: This code is for "insert item", which looks at our snippets which are ARM resources to populate its list, then calls vscode to insert the resource that the user selects.  Now instead goes through the snippet manager

src/insertItem.ts
```diff
@@ -425,11 +403,11 @@ export class InsertItem {
                 prepend = `,\r\n\t\t`;
             }
         }
-        const resource = await this.ui.showQuickPick(getResourceSnippets(), { placeHolder: 'What resource do you want to insert?' });
+        const resource = await this.ui.showQuickPick(await getResourceSnippets(), { placeHolder: 'What resource do you want to insert?' });
         await this.insertText(textEditor, index, prepend);
         let newCursorPosition = this.getCursorPositionForInsertResource(textEditor, index, prepend);
         textEditor.selection = new vscode.Selection(newCursorPosition, newCursorPosition);
-        await commands.executeCommand('editor.action.insertSnippet', { name: resource.label });
+        await commands.executeCommand('editor.action.insertSnippet', { snippet: resource.data.insertText });
```
StephenWeatherford, line 410: Now we we insert the snippet by text rather than name since vscode no longer knows about them.

test/functional/snippets.test.ts
```diff
@@ -416,53 +422,42 @@ suite("Snippets functional tests", () => {
...
+        let editor = await window.showTextDocument(doc);
 
         // Wait for first set of diagnostics to finish.
         await getDiagnosticsForDocument(doc, {});
-        const initialDocText = window.activeTextEditor!.document.getText();
+        const initialDocText = doc.getText();
 
-        // Start waiting for next set of diagnostics (so it picks up the current completion versions)
-        let diagnosticsPromise: Promise<Diagnostic[]> = getDiagnosticsForDocument(
+        // Insert snippet (and wait for and verify diagnotics)
```
bhsubra, line 431: nit: typo

Issue #770, opened 2020-06-09
Feature: Code lenses, errors, intellisense and code actions for parameters of nested templates (generalized features from parameter files)


StephenWeatherford:
## TODOs
- [x] Clicking on value in top-level parameters goes to param file
- [x] Code lens for values for nested inner scoped templates
- [x] Clicking on value in nested template goes to template file
- [x] fix keyvault parameters
- [x] Add missing parameters from f1 not working
- [x] Clicking on keyvault value doesn't go to right place -> https://github.com/microsoft/vscode-azurearmtools/issues/876
- [x] Parameter file behavior in nested templates
  - [x] error param missing
  - [x] code actions
  - [x] Click on "no value found" when no parameters property
- [x] testcases
- [x] Parameter source: parameter file
  - [x] Error loading param file?
  - [x] Error: missing params
  - - [x] Quick fix
  - [x] Add missing parameters
    - [x] What if no "properties"?
  - [x] Top-level parameters
  - - [x] Code lens: current params file
  - - - [x] Click to navigate
  - - [x] Code lens: change params file
  - - [x] Code lens: no param file
  - - [x] Code lens: parameter sources
  - - [x] Code lens: click on parameter source (value, default, not found)
  - - - [x] Error loading param file?
  - - [x] Intellisense for parameter values
- [x] Parameter source: nested deployment
  - [x] Error loading param file?
  - [x] Error: missing params
  - - [x] Quick fix
  - [x] Add missing parameters
    - [x] What if no "properties"?  
Error Message: _span's length must be greater than or equal to 1.
new Issue Language.js:253:9
Object.getMissingParameterErrors getMissingParameterErrors.js:18:13
DeploymentTemplate.getMissingParameterErrors DeploymentTemplate.js:244:65
  - - [x] Code lens: parameter values
  - - [x] Code lens: click on parameter source (value, default, not found)
  - - [x] Intellisense for parameter values
- [ ] Telemetry

PR #826, opened 2020-07-06
Abstract param defs/vars in preparation for 770

StephenWeatherford:
This refactoring is the main work needed to implement https://github.com/microsoft/vscode-azurearmtools/issues/770. It abstracts the idea of parameter definitions and parameter values so that code lenses, code actions and Intellisense can be applied outside of the top-level parameter definitions of a template file and the parameter values specified in a parameter file so that they could also be provided for nested or linked deployments.
 
There are no (major) functionality changes yet, just making sure everything works as currently, the actual nested/linked support will be added in a separate PR.


StephenWeatherford:
Thanks!

PR #894, opened 2020-07-31
Fix 891, warnings not showing for parameter values for outer-scoped n…

StephenWeatherford:
…ested template

Fixes #894 

We were incorrectly treating the parameters object as an array.  Not sure how that happened, since the testcase I created for it had it working correctly in the screenshot.  Oh, well...


StephenWeatherford:
thx

PR #954, opened 2020-09-11
Add list() function

### LINKED ISSUE #949: List function is not recognized
List function is not recognized by extension although it works correctly in ARM Template
image

StephenWeatherford:
Fixes #949

### DIFF
--- a/assets/ExpressionMetadata.json
+++ b/assets/ExpressionMetadata.json
@@ -332,7 +332,7 @@
         {
             "name": "list",
             "expectedUsage": "list(resourceName\/resourceIdentifier, apiVersion)",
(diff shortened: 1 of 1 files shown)


assets/ExpressionMetadata.json
```diff
@@ -329,6 +329,12 @@
             "minimumArguments": 2,
             "maximumArguments": 2
         },
+        {
+            "name": "list",
+            "expectedUsage": "list(resourceName\/resourceIdentifier, apiVersion)",
+            "description": "Performs a GET call to a REST API on the given resource. The returned data depends on the resource type. See https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-resource#valid-uses.",
```
bmoore-msft, line 335: Should we say?

... The function arguments and the returned data depend on the resource type...

Technically true for all of them, but this is likely to be the most common/ambiguous.

bmoore-msft (approved):

bhsubra (approved):

Issue #1180, opened 2021-02-11
Fails to download dotnet if dotnet is installed but framework not found (e.g. 5.0 only)


JustinGrote:
## Issue
If you have dotnet installed but not the correct framework, the extension will fail to start and you must download the runtime manually.

## Recommendation
Integrate the MS dotnet downloader extension to automatically detect and use a common runtime of other dotnet-based extensions to avoid bloat.
https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.vscode-dotnet-runtime

## Log
```
Using dotnet core executable at C:\Users\jgrote\AppData\Local\Microsoft\dotnet\dotnet.exe
Starting ARM Template Language Server at c:\Users\JGrote\.vscode-insiders\extensions\msazurermtools.azurerm-vscode-tools-0.14.1\languageServer\Microsoft.ArmLanguageServer.dll
Language server nuget version: 3.0.0-preview.21058.5
Client options:
{
  "documentSelector": [
    {
      "language": "arm-template",
      "scheme": "file"
    },
    {
      "language": "arm-template",
      "scheme": "untitled"
    }
  ],
  "diagnosticCollectionName": "ARM Template Language Server diagnostics",
  "outputChannel": {
    "outputChannel": {
      "name": "Azure Resource Manager Tools",
      "extensionPrefix": "azureResourceManagerTools",
      "_outputChannel": {
        "name": "Azure Resource Manager Tools",
        "_channel": {}
      }
    }
  },
  "revealOutputChannelOn": 3,
  "synchronize": {
    "configurationSection": "azureResourceManagerTools"
  }
}
Server options:
{
  "run": {
    "command": "C:\\Users\\jgrote\\AppData\\Local\\Microsoft\\dotnet\\dotnet.exe",
    "args": [
      "c:\\Users\\JGrote\\.vscode-insiders\\extensions\\msazurermtools.azurerm-vscode-tools-0.14.1\\languageServer\\Microsoft.ArmLanguageServer.dll",
      "--logLevel",
      "Warning"
    ],
    "options": {
      "shell": false
    }
  },
  "debug": {
    "command": "C:\\Users\\jgrote\\AppData\\Local\\Microsoft\\dotnet\\dotnet.exe",
    "args": [
      "c:\\Users\\JGrote\\.vscode-insiders\\extensions\\msazurermtools.azurerm-vscode-tools-0.14.1\\languageServer\\Microsoft.ArmLanguageServer.dll",
[...]

PR #1181, opened 2021-02-11
Add warning when using dotnet path override

StephenWeatherford:


StephenWeatherford:
This is to help diagnose issues such as https://github.com/microsoft/vscode-azurearmtools/issues/1180.

PR #1243, opened 2021-03-06
Add warning for bad relativePath api version

### LINKED ISSUE #1242: Warning to change deployments resource apiVersion to "2020-06-01" for template specs, "2020-10-01" for direct deployment
> t’s easy to miss that you need to update the API version on the Deployments to "2020-10-01", unless you read the docs 😊… so perhaps a prompt message to promote updating, similar to the main schema prompt ?

StephenWeatherford:
Fixes #1242


StephenWeatherford:
Turns out 2020-06-01 is valid for template specs, and there's no way for me to detect when it's for a template spec.  Earlier api versions get an error for relativePath.

PR #1338, opened 2021-07-17
Add sleutho.tmcolor to vscode recommended extensions

asears:
As part of reviewing the arm-colors.tmcolor file, I added this recommended extension.


StephenWeatherford:
Thanks for the PR, but colorization hasn't changed since the original implementation, and the instructions in the code say what to do if needed.  This would make the recommendation every single time someone loads the project.

PR #1582, opened 2023-04-13
tslint -> eslint

StephenWeatherford:


AbhitejJohn:
@StephenWeatherford : Can we get a little more context on what this addresses? That would make a review easier since there's a quite a lot of file updates.

PR #1598, opened 2023-06-20
Adding Codeql task

AbhitejJohn:


AbhitejJohn:
Validating if this injects the CodeQL task correctly.

AbhitejJohn:
Nevermind - it looks like it was indeed run. I was just looking at the wrong build.

PR #1626, opened 2023-08-24
update snippet api versions

### LINKED ISSUE #1608: Snippets are using old API versions
Azure Resource Manager (ARM) Tools extension for Visual studio code v0.15.10 IntelliSense capability which populates the Azure resource schema generates the templates with API versions that are obsolete.
image

For e.g. here for both Storage accounts and Web Apps there are newer API versions than 2020-12-01 available with more features, this requires lot of manual additions of properties that otherwise could auto populate.
It's not something breaking but a suggestion to consider the feasibility of having new version of APIs integrated with the latest version of the ARM extension.

StephenWeatherford:
Updated the instructions for running the script, the rest is the result of running the automation.

fixes #1608


jfleisher (approved):
Other than that failing test this looks straightforward.

PR #1635, opened 2023-08-26
fix snippet tests

StephenWeatherford:
No idea how this change got through tests.


StephenWeatherford:
Actually doesn't block tests because they're run in non-debug mode so "debugger" doesn't stop the tests.

jfleisher (approved):

PR #1701, opened 2024-05-15
Fix current return values of tenant()

StephenWeatherford:


ocallesp:
How about changing 'country' with 'region' ?

ocallesp (approved):

PR #1734, opened 2024-07-17
Update changelog for 0.15.12

StephenWeatherford:


AbhitejJohn:
I'm assuming we are replacing instead of adding since we did not release November last year?

AbhitejJohn (approved):

PR #1760, opened 2024-09-11
Update maintenance mode notice

### LINKED ISSUE #1768: Update maintenance mode message
The ARM Tools extension is now in Maintenance Mode. That means that we will continue to support currently implemented features and capabilities, but we will not be able to prioritize adding new features. To take advantage of the best experience for deploying to Azure with infrastructure-as-code, we recommend switching to Bicep. Please note, that while this VS Code extension is in maintenance mode, ARM Template deployments as a whole are still fully supported and new features are still being added to the ARM Template language and runtime to support new capabilities in Bicep. If you wish to use these same capabilities directly in Bicep, they will be supported from a runtime perspective, but not in the ARM Tools VS Code extension.

StephenWeatherford:
Fixes #1768

image


puicchan (approved):
LGTM

AbhitejJohn (approved):

PR #1765, opened 2024-10-07
Add weekly build schedule to Azure Pipelines

StephenWeatherford:
Just to catch any issues when they occur

### DIFF
--- a/.azure-pipelines/main.yml
+++ b/.azure-pipelines/main.yml
@@ -8,6 +8,14 @@ resources:
 variables:
   Codeql.Enabled: true
 
+schedules:
+  - cron: "30 14 * * 0" # This sets the schedule to run weekly on Sundays at 2:30 PM
+    displayName: Weekly build
+    branches:
+      include:
+        - main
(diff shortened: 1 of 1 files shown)


.azure-pipelines/main.yml
```diff
@@ -8,6 +8,14 @@ resources:
 variables:
   Codeql.Enabled: true
 
+schedules:
```
AbhitejJohn, line 11: Can we only run this when there are actually new changes to build?

AbhitejJohn (approved):

.azure-pipelines/main.yml
StephenWeatherford: That's how we end up with broken pipelines, it's happened so many times in the past.  I have most of the others running regularly.  If there's anything you suggest changing, like retention policy, etc., let me know, but we need a test build every so often.

AbhitejJohn: Sounds good, lets take it through then.

PR #1783, opened 2024-10-31
Preparation for release

StephenWeatherford:


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "azurerm-vscode-tools",
     "displayName": "Azure Resource Manager (ARM) Tools",
     "description": "Language server, editing tools and snippets for Azure Resource Manager (ARM) template files.",
-    "version": "0.15.13-alpha",
```
AbhitejJohn, line 5: I'm a little confused on the removal of "alpha" in the version here. Seems like we've [added it](https://github.com/microsoft/vscode-azurearmtools/commit/041c3c06251bac5898601433516980c842451fa2) earlier in the year. Curious, how we handle versioning - I'm definitely missing context from earlier.

AbhitejJohn (approved):
Minor clarification on the versioning strategy. Everything else looks good to me.

package.json
StephenWeatherford: Preview is handled as you thought via `"preview": true`.  I use "-alpha" for all non-production builds, so that telemetry which comes from me or our testing team off of main is distinguishable from customer data.  After release I'll update to next version and add -alpha until time to release again.

